PR #31, opened 2020-08-05
meson.build: force finding an absolute path for cli/test.sh

delroth:
Currently meson ends up resolving the script path to the literal string
"cli/test.sh", which breaks when trying to run the tests from a
different directory than the source directory. A common reason one might
want to do this is if running the tests via "ninja test" instead of
"meson test".

TESTED=Tests successfully run with "ninja test" with this change.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -37,7 +37,7 @@ glome_test = executable('glome_test', 'glome_test.c', dependencies : glib_dep,
   link_with : glome_lib, include_directories : glome_incdir)
 test('glome', glome_test)
 
-cli_test = find_program('cli/test.sh')
+cli_test = find_program(meson.current_source_dir() + '/cli/test.sh')
 test('cli', cli_test,
   args : glome_cli,
   timeout : 120)


delroth:
Do you have a source that documents it as deprecated? I understand it's a wrapper, and it is almost certainly not as featureful, but it's also a fairly well established convention to be able to run tests through the build system as opposed to the build file generator (which is just plain weird, tbf). Dating back all the way to autotools and "make check".

pkern:
It's the only way documented for running them on https://mesonbuild.com/Unit-tests.html. Also my understanding is that with `meson test` you get additional flags like "actually show me the error log from the test".

It's not that I don't agree with you that it's weird that you call out to ninja to build and meson to test, fwiw. But my primary question also was if that's working correctly when done from the appropriate subdir. ;)

delroth:
Ah, right. I unfortunately don't have the answer to your question, since this is literally the first time I touch a Meson build file :-) The workaround I used here comes from https://github.com/mesonbuild/meson/issues/374 which deals with a different problem.

(and re: the other discussion branch, "ninja test" calls out to "meson test --print-errorlogs" by default, unless errorlogs were explicitly disabled at configure time.)

pkern:
Could you recheck with the current code if that already resolved your concern?

pkern:
We think that this issue is resolved. Please reopen if you disagree.

PR #42, opened 2020-08-10
login: Describe lockdown mode

### LINKED ISSUE #26: glome-login/README.md missing
We should document the various options for glome-login and explain what e.g. lockdown mode is.

bluecmd:
Fixes #26

### DIFF
--- a/login/README.md
+++ b/login/README.md
@@ -9,6 +9,30 @@ replacement of login(1).
  1. Create a configuration file, see [example.cfg](example.cfg).
  1. Try it out by running `glome-login -c glome.cfg -- root`
 
+## Lockdown
+
+Lockdown mode allows you to restrict logins on machines.
+It is similar to nologin(5), except that it applies to all users
+that are trying to lock in.
+
+Example usage of lockdown is that you can allow logins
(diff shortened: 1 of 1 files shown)


bluecmd:
PTAL

login/README.md
```diff
@@ -9,6 +9,27 @@ replacement of login(1).
  1. Create a configuration file, see [example.cfg](example.cfg).
  1. Try it out by running `glome-login -c glome.cfg -- root`
 
+## Lockdown
```
pkern, line 12: I'd suggest to intro with what the purpose is, not how it works. Something like this maybe?

Lockdown mode allows you to restrict logins on machines. If the lockdown file exists and has the expected content, then logins will be prohibited. It is similar to `nologin(5)`, except that it applies to all users that are trying to lock in. It can be used to only allow logins on broken machines or machines in a certain state that is safe to log into and otherwise be certain that no person will be able to log in over the console.

And then you can talk about expected content. Although you also say "by default" and it's not really possible to change this today, as far as I know.

bluecmd: I have that in the Example usage, but I guess a one-two sentence introduction might be worth it. PTAL.

PR #45, opened 2020-08-20
Safer GLOME CLI

burgerdev:
This PR reworks the invocation of the `glome` utility to make it unambiguous in its arguments and overall safer to use correctly.

* private key and peer key are now always expected to be files
* arguments are passed as flags so that they cannot be passed in the wrong order accidentally
* messages (which may have sensitive content) are passed via `stdin` to reduce possibility of leaks
* the code is restructured to separate `main` concerns from the functionality

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,8 +54,8 @@ jobs:
           glome.c glome.h glome_test.c
     - name: check glome-cli format
       run: |
-        clang-format --Werror --dry-run --style=google cli/main.c
-
+        clang-format --Werror --dry-run --style=google \
+          cli/*.c cli/*.h
     - name: check glome-login format
       run: |
         clang-format --Werror --dry-run --style=google \
--- a/cli/commands.c
+++ b/cli/commands.c
@@ -1,0 +1,230 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "commands.h"
+
+#include <errno.h>
+#include <error.h>
+#include <fcntl.h>
+#include <getopt.h>
+#include <limits.h>
+#include <openssl/crypto.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "glome.h"
+
+#define GLOME_CLI_MAX_MESSAGE_LENGTH 4095
+
+// Arguments
+static const char *key_file = NULL;
+static const char *peer_file = NULL;
+static const char *tag_hex = NULL;
+static unsigned long counter = 0;
+
+static bool parse_args(int argc, char **argv) {
+  struct option long_options[] = {{"key", required_argument, 0, 'k'},
+                                  {"peer", required_argument, 0, 'p'},
+                                  {"counter", required_argument, 0, 'c'},
+                                  {"tag", required_argument, 0, 't'},
+                                  {0, 0, 0, 0}};
+
+  int c;
+  while ((c = getopt_long(argc, argv, "c:k:p:t:", long_options, NULL)) != -1) {
+    switch (c) {
+      case 'c': {
+        char *endptr;
+        counter = strtoul(optarg, &endptr, 0);
+        if (counter > UCHAR_MAX || optarg[0] == '\0' || *endptr != '\0') {
+          fprintf(stderr, "'%s' is not a valid counter (0..255)\n", optarg);
+          return false;
+        }
+        break;
+      }
+      case 'k':
+        key_file = optarg;
+        break;
+      case 'p':
+        peer_file = optarg;
+        break;
+      case 't':
+        tag_hex = optarg;
+        break;
+      case '?':
+        return false;
+      default:
+        // option not implemented
+        abort();
+    }
+  }
+  return true;
+}
+
+static int decode_hex(uint8_t *dst, size_t dst_len, const char *in) {
+  size_t len = strlen(in);
+  if (len > 2 && in[0] == '0' && in[1] == 'x') {
+    len -= 2;
+    in += 2;
+  }
+  if (len > dst_len * 2 || len % 2 != 0) {
+    return -1;
+  }
+  size_t i;
+  for (i = 0; i < len / 2; i++) {
+    if (sscanf(in + (i * 2), "%02hhX", dst + i) != 1) {
+      fprintf(stderr, "ERROR while parsing byte %zu ('%c%c') as hex\n", i,
+              in[2 * i], in[2 * i + 1]);
+      return -3;
+    }
+  }
+  return i;
+}
+
+static bool read_file(const char *fname, uint8_t *buf, const size_t num_bytes) {
+  FILE *f = fopen(fname, "r");
+  if (!f) {
+    fprintf(stderr, "could not open file %s: %s\n", fname, strerror(errno));
+    return false;
+  }
+  if (fread(buf, 1, num_bytes, f) != num_bytes) {
+    fprintf(stderr, "could not read %zu bytes from file %s", num_bytes, fname);
+    if (ferror(f)) {
+      fprintf(stderr, ": %s\n", strerror(errno));
+    } else {
+      fputs("\n", stderr);
+    }
+    fclose(f);
+    return false;
+  }
+  fclose(f);
+  return true;
+}
+
+static void print_hex(FILE *stream, const char *prefix, uint8_t *buf,
+                      size_t len) {
+  if (prefix != NULL) {
+    fputs(prefix, stream);
+  }
+  for (size_t i = 0; i < len; i++) {
+    fprintf(stream, "%02x", buf[i]);
+  }
+  fputs("\n", stream);
+}
+
+int genkey(int argc, char **argv) {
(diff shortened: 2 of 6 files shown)


cli/test.sh
```diff
@@ -33,19 +33,23 @@ t=$(mktemp -d glometest.XXX)
 trap "rm -rf -- '${t?}'" EXIT
```
pkern, line 34: Maybe for a follow-up CL I think it'd be worthwhile to actually validate the test vectors.

This worked for me:

```python
>>> open('PRIVATE-KEY', mode='wb').write(binascii.unhexlify('77076d0a7318a57d3c16c17251b26645df4c2f87ebc0992ab177fba51db92c2a'))
32
>>> open('PUBLIC-KEY', mode='wb').write(binascii.unhexlify('de9edb7d7b7dc1b4d35b61c2ece435373f8343c85b78674dadfc7e146f882b4f'))
32
```
```shell
$ echo -n "The quick brown fox" | ./glome tag --key PRIVATE-KEY --peer PUBLIC-KEY 
9c44389f462d35d0672faf73a5e118f8b9f5c340bbe8d340e2b947c205ea4fa3
```

cli/commands.c
```diff
@@ -0,0 +1,230 @@
...
+  for (size_t i = 0; i < len; i++) {
+    fprintf(stream, "%02x", buf[i]);
+  }
+  fputs("\n", stream);
+}
+
+int genkey(int argc, char **argv) {
+  uint8_t private_key[GLOME_MAX_PRIVATE_KEY_LENGTH] = {0};
+
+  // TODO: maybe verify stdout umask 077
```
pkern, line 131: What does that mean?

cli/main.c
```diff
@@ -23,157 +24,51 @@
...
-    print_hex(stdout, NULL, tag, sizeof tag);
+  if (argc < 2) {
+    return print_help(argc, argv);
   }
-
-  return EXIT_SUCCESS;
+  // Advance to the requested subcommand.
+  const struct cmd *c;
+  for (c = cmds; c->name && strcmp(c->name, argv[1]); c++)
+    ;
```
pkern, line 72: That's almost "smart" code. Could you add a comment to `cmds[]` that the last one will be executed when none is found? Also I think you always need to start a block, although I'm somewhat unsure about this. I almost feel like a while loop might be easier to read.

cli/commands.c
```diff
@@ -0,0 +1,230 @@
...
+int tag(int argc, char **argv) {
+  uint8_t tag[GLOME_MAX_TAG_LENGTH] = {0};
+  if (!parse_args(argc, argv)) {
+    return EXIT_FAILURE;
+  }
+  if (!key_file || !peer_file) {
+    fprintf(stderr, "not enough arguments for subcommand %s\n", argv[1]);
+    return EXIT_FAILURE;
+  }
+  int res = tag_impl(tag, /* verify */ false, key_file, peer_file);
```
pkern, line 198: `/*verify=*/ false`

cli/commands.c
```diff
@@ -0,0 +1,230 @@
...
+  uint8_t expected_tag[GLOME_MAX_TAG_LENGTH] = {0};
+  size_t expected_tag_len = 0;
+  if (!parse_args(argc, argv)) {
+    return EXIT_FAILURE;
+  }
+  if (!key_file || !peer_file || !tag_hex) {
+    fprintf(stderr, "not enough arguments for subcommand %s\n", argv[1]);
+    return EXIT_FAILURE;
+  }
+  int res = tag_impl(tag, /* verify */ true, key_file, peer_file);
```
pkern, line 217: Same as above.

cli/commands.c
```diff
@@ -0,0 +1,230 @@
...
+  if (res) {
+    return res;
+  }
+
+  // compare the tag
+  expected_tag_len = decode_hex(expected_tag, sizeof expected_tag, tag_hex);
+  // TODO: should this be constant time?
```
pkern, line 224: Yes.

pkern (changes requested):

cli/commands.c
burgerdev: In theory, we could verify that the file associated with stdout - *if there is a file* - is not world readable. wireguard does that. However, this does not help detecting an insecure `glome genkey | tee /tmp/foo.key`, so I'm inclined to just dropping the topic.

cli/test.sh
burgerdev: Opened #50 for it.

cli/commands.c
burgerdev: Replaced it with `CRYPTO_memcmp`.

cli/main.c
burgerdev: Is that a praise or an insult? :D Added comments, and it is much more readable with a while loop.

cli/commands.c
```diff
@@ -0,0 +1,230 @@
...
+        tag_hex = optarg;
+        break;
+      case '?':
+        return false;
+      default:
+        // option not implemented
+        abort();
+    }
+  }
+  return true;
```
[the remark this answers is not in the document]
burgerdev: Adopted that suggestion, makes a bit more sense that way. Interestingly, I never saw the compiler complaining about the dead code.

.github/workflows/ci.yml
```diff
@@ -54,7 +54,8 @@ jobs:
           glome.c glome.h glome_test.c
     - name: check glome-cli format
       run: |
-        clang-format --Werror --dry-run --style=google cli/main.c
+        clang-format --Werror --dry-run --style=google \
+          cli/main.c cli/commands.c cli/commands.h
```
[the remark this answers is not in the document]
burgerdev: That was an explicit decision for dependencies, iirc. We should discuss whether it makes sense to keep it for formatting. The newly added `yapf` and `pylint` use wildcards, fwiw.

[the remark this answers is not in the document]
pkern: It wasn't as much a decision there: meson requires this for all inputs to ensure it can write out an accurate build graph for ninja. For formatting I'd say file endings would be fine. I'd go as far as **/*.c and **/*.h all need to be formatted correctly. Obviously that breaks down once you have the first exception, though. ;-)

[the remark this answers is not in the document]
burgerdev: Switched to using a wildcard for formatting.

cli/commands.c
```diff
@@ -0,0 +1,228 @@
...
+    fprintf(stderr, "could not read %zu bytes from file %s", num_bytes, fname);
+    if (ferror(f)) {
+      fprintf(stderr, ": %s\n", strerror(errno));
+    } else {
+      fputs("\n", stderr);
+    }
+    fclose(f);
+    return false;
+  }
+  return true;
```
pkern, line 114: Missing fclose?

cli/commands.c
```diff
@@ -0,0 +1,228 @@
...
+  }
+  fputs("\n", stream);
+}
+
+int genkey(int argc, char **argv) {
+  uint8_t private_key[GLOME_MAX_PRIVATE_KEY_LENGTH] = {0};
+
+  if (glome_generate_key(private_key, NULL)) {
+    fprintf(stderr, "unable to generate a new key\n");
+  }
```
pkern, line 133: Missing return?

cli/commands.c
```diff
@@ -0,0 +1,228 @@
...
+// direction determined by verify.
+int tag_impl(uint8_t tag[GLOME_MAX_TAG_LENGTH], bool verify,
+             const char *key_file, const char *peer_file) {
+  uint8_t private_key[GLOME_MAX_PRIVATE_KEY_LENGTH] = {0};
+  uint8_t peer_key[GLOME_MAX_PUBLIC_KEY_LENGTH] = {0};
+  char message[GLOME_CLI_MAX_MESSAGE_LENGTH] = {0};
+
+  if (!read_file(key_file, private_key, sizeof private_key) ||
+      !read_file(peer_file, peer_key, sizeof peer_key)) {
+    return EXIT_FAILURE;
```
pkern, line 172: No perror here?

cli/main.c
```diff
@@ -23,157 +24,55 @@
...
-    if (sscanf(in + (i * 2), "%02hhX", dst + i) != 1) {
-      fprintf(stderr, "ERROR while parsing byte %zu ('%c%c') as hex\n", i,
-              in[2 * i], in[2 * i + 1]);
-      return -3;
-    }
-  }
-  return i;
-}
+    "    %s verify --key PRIVATE-KEY --peer PEER-KEY --tag TAG [--counter "
+    "COUNTER] <MESSAGE\n";
```
pkern, line 42: `<MESSAGE` feels weird to me. `echo MESSAGE |`? Or `<MESSAGE>`?

cli/main.c
```diff
@@ -23,157 +24,55 @@
...
-  uint8_t public_key[GLOME_MAX_PUBLIC_KEY_LENGTH] = {0};
-  uint8_t peer_key[GLOME_MAX_PUBLIC_KEY_LENGTH] = {0};
-  uint8_t tag[GLOME_MAX_TAG_LENGTH] = {0};
-  uint8_t expected_tag[GLOME_MAX_TAG_LENGTH] = {0};
-  size_t expected_tag_len = 0;
-  char *message = NULL;
-  char *endptr;
-  bool verification = false;
-  unsigned long counter = 0;
+typedef int (*mainfunc)(int, char **);
```
pkern, line 49: Would it be legit to name parameters in a func typedef?

pkern (changes requested):

cli/commands.c
burgerdev: `read_file` already prints debug info.

cli/main.c
burgerdev: I added `-FILE` suffixes to make the nature of these arguments more obvious. I'd use `<MESSAGE>` in a man page synopsis, but rather not in this example invocation to avoid confusion with shell redirections.

cli/main.c
burgerdev: Yes, the rules seem to be the same as for function declarations - added names.

pkern (approved):
LGTM.

I still have some concerns around newlines, i.e. the fact that you need `printf` to ensure that the message you pass in does not have one. That's still super annoying if you actually craft a file using a text editor. But alas, I don't see a great way of fixing that and it's now just part of the calling convention.

PR #46, opened 2020-08-20
Go implementation of GLOME

pedrobn23:
This is a draft PR to show the progress in the GLOME golang library. Any review/comment are welcome.


go/glome/glome.go
```diff
@@ -0,0 +1,194 @@
...
+	// PublicKeySize is the size of a PublicKey in bytes.
+	PublicKeySize = 32
+	// SharedSecretSize is the size of a SharedSecret in bytes.
+	SharedSecretSize = 32
+	// Maximum size allow for a Tag
+	MaxTagSize = 32
+)
+
+var (
+	errInvalidPublicKey = errors.New(
```
Republicn, line 39: Maybe InvalidPublicKeyErr(or)? Sounds more natural.

go/glome/glome.go
```diff
@@ -0,0 +1,194 @@
...
+		fmt.Sprintf("glome: invalid public key - byte slice len is not %d", PublicKeySize))
+	errInvalidPrivateKey = errors.New(
+		fmt.Sprintf("glome: invalid private key - byte slice len is not %d", PrivateKeySize))
+	errInvalidTagSize = errors.New(
+		fmt.Sprintf("glome: invalid tag size - MinTagSize must be in range 1-%d", MaxTagSize))
```
Republicn, line 44: `1` is a constant too, so you may, for example, save it as `MinPossibleTagSize`.

go/glome/glome.go
```diff
@@ -0,0 +1,194 @@
...
+	"fmt"
+	"golang.org/x/crypto/curve25519"
+	"io"
+)
+
+const (
+	// PrivateKeySize is the size of a PrivateKey in bytes.
+	PrivateKeySize = 32
+	// PublicKeySize is the size of a PublicKey in bytes.
+	PublicKeySize = 32
```
Republicn, line 31: Could it be that private and public keys have different sizes?

go/glome/glome.go
pedrobn23: IMHO starting error variables with "err" is quite idiomatic (https://github.com/golang/go/wiki/Errors#naming).

go/glome/glome.go
pedrobn23: In this implementation with the standard GLOME protocol variation is not possible. In other protocols (or in other implementations) it is possible that public/private keys have different values (e.g. https://github.com/golang/go/blob/master/src/crypto/ed25519/ed25519.go)

go/glome/glome.go
pedrobn23: Fixed.

go/README.md
```diff
@@ -0,0 +1,69 @@
...
+```go
+ex := alicePriv.Exchange(bobPub)
+firstTag = ex.Tag(msg, 0)
+```
+
+And Alice will send Bob both `msg`, `firstTag` as well as Alice's public key.
+On Bob ends he will need to do the following:
+
+```go
+ex := alicePriv.Exchange(bobPub)
```
burgerdev, line 53: Should this be `bobPriv.Exchange(alicePub)`?

pedrobn23: Yes, it should be! fixed.

go/glome/glome.go
```diff
@@ -0,0 +1,199 @@
...
+		fmt.Sprintf("glome: invalid public key - byte slice len is not %d", PublicKeySize))
+	ErrInvalidPrivateKey = errors.New(
+		fmt.Sprintf("glome: invalid private key - byte slice len is not %d", PrivateKeySize))
+	ErrInvalidTagSize = errors.New(
+		fmt.Sprintf("glome: invalid tag size - MinPeerTagSize must be in range %d-%d",
+			MinTagSize, MaxTagSize))
+)
+
+// SharedSecret is the type of GLOME (and x25519) shared secret.
+type SharedSecret [SharedSecretSize]byte
```
pkern, line 53: I have not been a fan of fixed size byte slices and I have rejoiced when `curve25519.X25519` was added, which takes `[]byte` instead of `[32]byte`. The latter are just unwieldy to work with, as you need `[:]` everywhere and need to copy things.

The other question is if you want to make the type opaque. If so (and I'm not sure you want to), you'd do `type SharedSecret struct{ secret []byte }`, which makes the member inaccessible.

go/glome/glome.go
```diff
@@ -0,0 +1,199 @@
...
+
+// PublicKey is the type of GLOME (and x25519) public Keys.
+type PublicKey [PublicKeySize]byte
+
+// Generate a PublicKey object from slice. Return ErrInvalidPublicKey
+// if slice's length is not PublicKeySize.
+func PublicKeyFromSlice(b []byte) (*PublicKey, error) {
```
pkern, line 60: I'd tend to return byte slices as that, not as pointers. Especially as you don't want anyone to modify where the byte slice points and your byte slices are actually ephemeral (i.e. newly generated within the function).

go/glome/glome.go
```diff
@@ -0,0 +1,199 @@
...
+	if len(b) != PublicKeySize {
+		return nil, ErrInvalidPublicKey
+	}
+
+	var p PublicKey
+	copy(p[:], b)
+	return &p, nil
+}
+
+// PrivateKey is the type of GLOME (and x25519) shared secret.
```
pkern, line 70: That's a weird statement. A PrivateKey is usually not a shared secret. If it's used for both shared secret and private key, then write that. But SharedSecret actually has its own type already.

go/glome/glome.go
```diff
@@ -0,0 +1,199 @@
...
+
+	var s SharedSecret
+	copy(s[:], slice)
+	return &Dialog{shared: s, User: *public, Peer: *peer, MinPeerTagSize: MaxTagSize}, nil
+}
+
+// Generates a Dialog struct. It performs GLOME handshake, and stores create a Dialog
+// from the user to the peer. Sets param m as MinPeerTagSize.
+func (user *PrivateKey) TruncatedExchange(peer *PublicKey, m uint) (*Dialog, error) {
+
```
pkern, line 105: stray empty line

go/glome/glome.go
```diff
@@ -0,0 +1,199 @@
...
+// in the context of a communication from the users to theirs peers.
+func (d *Dialog) Tag(msg []byte, c uint8) []byte {
+	return tag(msg, c, d.sendingKey())
+}
+
+// Checks if a tag matches some provided message and counter.
+// The method generates the matching tag following GLOME protocol
+// specification in the context of a communication from the users'
+// peers to the users and then is compared with the tag provided.
+func (d *Dialog) Check(t []byte, msg []byte, c uint8) bool {
```
pkern, line 186: Method parameters in Go should not be short, as they are part of the documentation. I'd suggest to s/t/tag/ and s/c/counter/ everywhere. s/msg/message/ is debatable but might make sense for clarity.

go/glome/glome.go
```diff
@@ -0,0 +1,199 @@
...
+	switch {
+	case uint(len(t)) < d.MinPeerTagSize:
+		prefixSize = d.MinTagSize
+	case uint(len(t)) > MaxTagSize:
+		prefixSize = MaxTagSize
+	default:
+		prefixSize = uint(len(t))
+	}
+
+	correct := tag(msg, c, d.receivingKey())[:prefixSize]
```
pkern, line 197: s/correct/want/ (mostly because tests would do that too)

pkern (changes requested):

go/glome/glome.go
pedrobn23: Fixed, it was an error.

go/glome/glome.go
pedrobn23: Fixed.

go/glome/glome.go
pedrobn23: Fixed.

go/glome/glome.go
pedrobn23: Fixed.

go/glome/glome.go
pedrobn23: Decided to delete this type.

go/glome/glome.go
```diff
@@ -0,0 +1,191 @@
...
+func (d *Dialog) sendingKey() []byte {
+	return append(d.shared[:], append(d.Peer[:], d.User[:]...)...)
+}
+
+func (d *Dialog) receivingKey() []byte {
+	return append(d.shared[:], append(d.User[:], d.Peer[:]...)...)
+}
+
+// Generates a tag matching some provided message, counter and password.
+func generate_tag(msg []byte, counter uint8, password []byte) []byte {
```
Republicn, line 160: shouldn't it be `generateKey`?

pedrobn23: Yes, it has an error. It should be generateTag.

pedrobn23: Fixed.

go/README.md
```diff
@@ -0,0 +1,69 @@
+# GLOME-Go
+**This is not an officially supported Google product.**
+
+This repository contains a Golang implementation for the GLOME protocol. You can
+find the library as well as the tests on the folder glome. 
+
+
+## Python API
```
Republicn, line 8:
```suggestion
## Go API
```

go/glome/glome_test.go
```diff
@@ -0,0 +1,159 @@
...
+	if e != nil {
+		t.Fatalf("Unexpected Error: " + e.Error())
+	}
+}
+
+// Stores test vectors from protocol reference. Each variable is named after
+// a row of the test table. For the purpose of testing, we consider that user A
+// is always the one that sends the message (therefore, we change the role of A
+// B in Vector #2)
+type test struct {
```
Republicn, line 33: testVector? as it's not itself a test

go/glome/glome.go
pedrobn23: The only thing this function does is to take a slice and transform it into a pointer to PublicKey, so that there is a good interaction between both types. If this function returns a slice:

- The PublicKey type functionality would not be accessible from the object it returns.
- The only functionality of this function would be to check if the last slice had 32bytes. It would be better in that case to return only the existence of error.

Another question apart from the type that returns this function is if PublicKey should be of type [32]byte or []byte. I chose the first one so that the library could avoid dealing with errors in time of use, and only add this function for when it was convenient for the user to have a slice.

go/glome/glome_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return sending, receiving
+}
+
+func decodeString(s string) []byte {
+	b, err := hex.DecodeString(s)
+	if err != nil {
+		return []byte{0} // We return a single byte.
+	}
+
```
pkern, line 68: I'd remove this stray newline.

go/glome/glome_test.go
```diff
@@ -0,0 +1,160 @@
...
+	receiving, err := bPriv.Exchange(aPub)
+	handle(err, t)
+
+	return sending, receiving
+}
+
+func decodeString(s string) []byte {
+	b, err := hex.DecodeString(s)
+	if err != nil {
+		return []byte{0} // We return a single byte.
```
pkern, line 66: Why? If this is invalid input in a test, then panic here instead of passing along corrupted data.

go/glome/glome_test.go
```diff
@@ -0,0 +1,160 @@
...
+			want  []byte
+		}{
+			{input: send.sendingKey(), want: append(tv.ks, append(tv.kb, tv.ka...)...)},
+			{input: send.receivingKey(), want: append(tv.ks, append(tv.ka, tv.kb...)...)},
+			{input: rec.sendingKey(), want: append(tv.ks, append(tv.ka, tv.kb...)...)},
+			{input: rec.receivingKey(), want: append(tv.ks, append(tv.kb, tv.ka...)...)},
+		}
+
+		t.Run(name, func(t *testing.T) {
+			for _, k := range keys {
```
pkern, line 113: Usually table-driven tests are defined inline:

```go
for _, k := []struct{
  in, want []byte
}{
  {case1},
  ...
} {
  ...
}
```

(Also below)

go/glome/glome_test.go
```diff
@@ -0,0 +1,160 @@
...
+			counter: 100,
+			msg:     []byte("The quick brown fox"),
+			ks:      decodeString("4b1ee05fcd2ae53ebe4c9ec94915cb057109389a2aa415f26986bddebf379d67"),
+			tag:     decodeString("06476f1f314b06c7f96e5dc62b2308268cbdb6140aefeeb55940731863032277"),
+		},
+	}
+}
+
+func TestKeyGeneration(t *testing.T) {
+	for name, tv := range tests() {
```
pkern, line 99: I'd capture `name := name` at the beginning of the loop for safety in case someone adds `t.Parallel()` later.

go/glome/glome_test.go
```diff
@@ -0,0 +1,160 @@
...
+			}
+		})
+	}
+}
+
+func TestTagGeneration(t *testing.T) {
+	for name, tv := range tests() {
+		send, _ := tv.Dialogs(t)
+		got := send.Tag(tv.msg, tv.counter)
+		if !bytes.Equal(tv.tag, got) {
```
pkern, line 126:
```go
if got := [...]; !bytes.Equal(...) { // i.e. one line
```

go/glome/glome.go
```diff
@@ -0,0 +1,223 @@
...
+	if m == 0 || m > MaxTagSize {
+		return nil, ErrInvalidTagSize
+	}
+
+	d, err := priv.Exchange(peer)
+	if err != nil {
+		return nil, err
+	}
+
+	d.MinPeerTagSize = m
```
pkern, line 126: You validate the field but you make it exported. That means I could go and do `d.MinPeerTagSize = 0` later. I think it'd be better to leave it unexported.

go/glome/glome.go
```diff
@@ -0,0 +1,223 @@
...
+
+	var p PrivateKey
+	copy(p[:], b)
+	return &p, nil
+}
+
+// GenerateKeys generates a public/private key pair using entropy from rand.
+func GenerateKeys(rand io.Reader) (*PublicKey, *PrivateKey, error) {
+	b := make([]byte, PrivateKeySize)
+	_, err := rand.Read(b)
```
pkern, line 145: Shouldn't you check that n from `n, err := rand.Read(b)` matches expectations? Lest you risk having truncated randomness in the byte array?

pkern (changes requested):

go/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Go API
+
+### Requirements
+
+- golang.org/x/crypto v0.0.0-20200820211705-5c72a883971a
+
+### Example
+
+We provide a brief example of use.  In order for Alice and Bob to communicate,
```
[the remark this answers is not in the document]
pedrobn23: Fixed.

go/README.md
```diff
@@ -0,0 +1,69 @@
...
+```go
+import (
+        "glome" 
+        "crypto/rand"
+)
+
+// Alice generates new random KeyPair
+alicePub, alicePriv, err := glome.GenerateKeys(rand.Reader)
+
+// Bob Generates Private Key from as existing byte array
```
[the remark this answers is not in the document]
pedrobn23: This is the default functionality provided in [crypto/ed25119](https://golang.org/pkg/crypto/ed25519/#GenerateKey) and is the defaul method for generating cryptographically secure random number in golang.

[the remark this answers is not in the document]
pedrobn23: Fixed

go/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+This repository contains a Golang implementation for the GLOME protocol. You can
+find the library as well as the tests on the folder glome. 
+
+
+## Go API
+
+### Requirements
+
+- golang.org/x/crypto v0.0.0-20200820211705-5c72a883971a
```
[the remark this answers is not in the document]
pedrobn23: Added.

go/README.md
```diff
@@ -0,0 +1,69 @@
...
+// Bob deduces public key
+bobPub := bobPriv.Public()
+```
+
+Suppose that Alice knows `bobPub` and wants to send Bob the message
+`msg` and no other message have been shared before. Alice will need to generate 
+a `Dialog`
+
+```go
+ex := alicePriv.Exchange(bobPub)
```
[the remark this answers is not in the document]
pedrobn23: I did this to show that the same Dialog can be used for more than one Tag. Will include more Tags to show that better. Also will change `ex` to `di`.

go/README.md
```diff
@@ -0,0 +1,69 @@
...
+tag := ex.Tag(msg, 0)
+```
+
+And Alice will send Bob both `msg`, `firstTag` as well as Alice's public key.
+On Bob ends he will need to do the following:
+
+```go
+ex := bobPriv.Exchange(alicePub)
+valid := ex.check(tag, msg, 0)
```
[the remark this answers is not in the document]
pedrobn23: Typo.

go/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+if !valid{
```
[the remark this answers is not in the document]
pedrobn23: Fixed.

go/README.md
```diff
@@ -0,0 +1,69 @@
...
+    // Maybe someone is pretending to be Alice!
+    // panic?
```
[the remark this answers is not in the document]
pedrobn23: Fixed.

go/README.md
```diff
@@ -0,0 +1,69 @@
...
+s := b[:]
+bobPriv, _ := glome.PrivateKeyFromSlice(s)
+
+// Bob deduces public key
+bobPub := bobPriv.Public()
+```
+
+Suppose that Alice knows `bobPub` and wants to send Bob the message
+`msg` and no other message have been shared before. Alice will need to generate 
+a `Dialog`
```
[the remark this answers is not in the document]
pedrobn23: Fixed.

go/glome/glome_test.go
pedrobn23: Fixed.

go/glome/glome.go
pedrobn23: Left it unexported.

go/glome/glome_test.go
pedrobn23: Fix.

go/glome/glome_test.go
```diff
@@ -0,0 +1,160 @@
...
+	kb      []byte
+	counter uint8
+	msg     []byte
+	ks      []byte
+	tag     []byte
+}
+
+func (tv *testVector) Dialogs(t *testing.T) (*Dialog, *Dialog) {
+	aPriv, err := PrivateKeyFromSlice(tv.kap)
+	handle(err, t)
```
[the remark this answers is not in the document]
pedrobn23: Fixed.

go/glome/glome.go
pedrobn23: Fixed.

go/glome/glome_test.go
pedrobn23: Fixed.

go/glome/glome.go
```diff
@@ -0,0 +1,228 @@
...
+	if len(b) != PublicKeySize {
+		return nil, ErrInvalidPublicKey
+	}
+
+	var p PublicKey
+	copy(p[:], b)
+	return &p, nil
+}
+
+// PrivateKey is the type of GLOME public Keys.
```
pkern, line 76:
```suggestion
// PrivateKey is the type of GLOME public keys.
```

go/glome/glome.go
```diff
@@ -0,0 +1,228 @@
...
+//
+// It can be initializes either by casting a [PrivateKeySize]byte array or from a byte
```
pkern, line 78:
```suggestion
// It can be initialized either by casting a [PrivateKeySize]byte array or from a byte
```

go/glome/glome.go
```diff
@@ -0,0 +1,228 @@
...
+// slice with the PrivateKeyFromSlice function. For example:
+//
+//     //Generate Private Key as existing byte array
+//     b := [32]byte{0,2,...,7,6}
+//     p := glome.PrivateKey(b)
+//     //Generate from byte slice
+//     s := b[:]
+//     p, _ := glome.PrivateKeyFromSlice(s)
```
pkern, line 86: So I remain unconvinced that this is a great API design. For the exchange I'd optimally want to do something like this:

```go
eph, err := glome.GenerateKey()
if err != nil { [...] }
// for glome-login we actually need an identifier, too
pub, err := ioutil.ReadFile("/etc/glome/service.pub")
if err != nil { [...] }
srv, err := glome.PublicKey(pub)
if err != nil { [...] }
tag, err := eph.Exchange(srv).Tag(msg)
if err != nil { [...] }
```

And the other side:

```go
// for glome-login we actually need an identifier, too
p, err := ioutil.ReadFile("/etc/glome/service.key")
if err != nil { [...] }
priv, err := glome.PrivateKey(p)
if err != nil { [...] }
// get peer public key
tag, err := priv.Exchange(peer).Tag(msg)
if err != nil { [...] }
```

It's definitely not ok to drop the error on the floor in the example when the whole point of the function is to do the length check. If you want to use byte arrays they should be an implementation detail.

pedrobn23: The API that we have now is really similiar to the one that you are requesting. For the exchange the actual code can be:
```go
_, eph, err := glome.GenerateKey(rand.Reader)
if err != nil { [...] }
// for glome-login we actually need an identifier, too
pub, err := ioutil.ReadFile("/etc/glome/service.pub")
if err != nil { [...] }
srv, err := glome.PublicKeyFromSlice(pub)
if err != nil { [...] }
tag, err := eph.Exchange(srv).Tag(msg)
if err != nil { [...] }
```

And the other side:
```go
// for glome-login we actually need an identifier, too
p, err := ioutil.ReadFile("/etc/glome/service.key")
if err != nil { [...] }
priv, err := glome.PrivateKeyFromSlice(p)
if err != nil { [...] }
// get peer public key
tag, err := priv.Exchange(peer).Tag(msg)
if err != nil { [...] }
```

All the difference are:
1. GenerateKey need a source of entropy. 
2. GenerateKey also returns the public Key - this can be change if it is considered that there will be no use for public key in most cases
3. `s/glome.PrivateKey(p)/glome.PrivateKeyFromSlice(p)` and
`s/glome.PublicKey(p)/glome.PublicKeyFromSlice(p)`

To address everyone of them:
1. This was decided in order not to impose the user which source of entropy to use. As there is no optional parameter in golang, to use a default allowing `nil` as a valid input was considered. It was discarded because it was regarded as unnecesary if how to generate it in a standard way (with `crypto/rand`) is conveniently documented.
2. I though that having a PublicKey from the begginning might be useful, but it is really a question of taste. Totally fine with changing this option if it is considered necessary
3. There is no difference between the sample and the actual code but the name of a function. We can discuss alternative for the naming.


Finally, I recall that this is not a great example by any means. Will change it to be more useful, both in documentation and in the README.

bluecmd: Discussed offline, let's move on by calling this current version of the API alpha and subject to change.

@pedrobn23 can you add a note somewhere in the README of the library that this is Alpha and the API is not fixed yet?

bluecmd (approved):
Alpha approved

go/glome/glome.go
burgerdev: This seems not to be resolved yet, and there is another instance of this typo further up.

pedrobn23: True, fixed!

PR #49, opened 2020-08-25
Pedrobn23/python ci

pedrobn23:
Add python-ci workflow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,3 +63,20 @@ jobs:
           login/base64.h login/config.h login/crypto.h login/lockdown.h \
           login/login.h login/ui.h
 
+  python_presubmit:
+    runs-on: ubuntu-18.04
+    container:
+      image: "python:3.6-buster"
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: install dependencies from PyPI
+      run: python -m pip install cryptography yapf pylint
+    - name: run yapf test
+      run: yapf -d --style google python/pyglome/*.py python/test/*.py
+    - name: run pylint test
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -63,3 +63,18 @@ jobs:
...
+    runs-on: ubuntu-18.04
+    container:
+      image: "python:3.6-buster"
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: install dependencies from PyPI
+      run: python -m pip install cryptography yapf pylint
+    - name: run unit tests
+      run: bash -c "cd python; python -m test"
```
burgerdev, line 76: As far as I understand, the value of `run` should already be exceuted with `sh -c`. Can you try removing the shell wrapper?

.github/workflows/ci.yml
```diff
@@ -63,3 +63,18 @@ jobs:
...
+    - name: run yapf test
+      run: yapf -d --style google python/pyglome/*.py python/test/*.py
+    - name: run pylint test
+      run: pylint python/pyglome/*.py
```
burgerdev, line 80: I would prefer to run lint and yapf before the tests, but it does hardly make a difference at this point. This would change if we introduced heavier integration tests, though.

.github/workflows/ci.yml
pedrobn23: Fixed.

.github/workflows/ci.yml
pedrobn23: Changed.

burgerdev (approved):

PR #52, opened 2020-08-27
Pedrobn23/documentation fixed

pedrobn23:
Improve documentation.

### DIFF
--- a/python/pyglome/glome.py
+++ b/python/pyglome/glome.py
@@ -72,6 +72,10 @@ class TagGenerationError(Error):
     """Raised whenever a tag could not be generated."""
 
 
+class ExchangeError(Error):
+    """Raised whenever the x25519 key exchange fails."""
+
+
 def _public_key_encode(public_key: x25519.X25519PublicKey):
     return public_key.public_bytes(serialization.Encoding.Raw,
                                    serialization.PublicFormat.Raw)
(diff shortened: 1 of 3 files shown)


python/pyglome/glome.py
```diff
@@ -114,7 +114,8 @@ def __init__(self,
               Must be an integer in range 1-32.
         Raises:
             ValueError: Raised whenever min_peer_tag_len is not in
-              range 1-32.
+              range 1-32. Also when null shared secret is derived from
+              user/peer key pair.
```
bluecmd, line 118: When does that happen? If it is the .exchange() function that returns null, consider using a specific error such as ExchangeError or something and actively check for None.

bluecmd (changes requested):

python/pyglome/glome.py
pedrobn23: Is the exchange function who raises ValueError exception when some public keys are provided. WDYT about `raise ExchangeError from ValueError`?

python/pyglome/glome.py
```diff
@@ -248,7 +257,8 @@ def __init__(self,
              documentation.
         Raises:
            ValueError: Raised whenever min_peer_tag_len is not in range 1-32 or
-             skippable_length is a negative integer.
+             skippable_length is a negative integer. Also when null shared secret
```
bluecmd, line 260: ExchangeError

bluecmd (changes requested):

python/pyglome/glome.py
pedrobn23: Fixed.

bluecmd (approved):

PR #53, opened 2020-09-09
Fixing prefix_tag_len check bug

Republicn:
There is a bug in login.c file, method request_url(...): it fails if prefix_tag_len == GLOME_MAX_TAG_LENGTH. This PR fixes it.

### DIFF
--- a/login/login.c
+++ b/login/login.c
@@ -144,7 +144,7 @@ int request_url(const uint8_t service_key[GLOME_MAX_PUBLIC_KEY_LENGTH],
                 const uint8_t prefix_tag[GLOME_MAX_TAG_LENGTH],
                 size_t prefix_tag_len, char** url, int* url_len,
                 const char** error_tag) {
-  if (prefix_tag_len >= GLOME_MAX_TAG_LENGTH) {
+  if (prefix_tag_len > GLOME_MAX_TAG_LENGTH) {
     return failure(EXITCODE_PANIC, error_tag, "prefix-tag-too-large");
   }
   // glome-handshake := base64url(


login/ui.h
```diff
@@ -55,10 +55,10 @@ typedef struct login_config {
   const char* config_file;
```
bluecmd, line 55: It would probably be better to rename config_file to config_path TBH. Path is less ambiguous than file ("Does this variable contain the file contents or the file path?").

bluecmd (changes requested):

login/ui.h
burgerdev: For comparison: the posix specification for the `exec` family uses `path` for variables that are directly used as a filesystem path, whereas `file` is used for variables that usually contain only a filename, i.e. for `PATH` lookup. https://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html. It would make sense to follow this convention at least for `login_path`, because it is passed to `execl` eventually. For consistency, I would then argue to use it for all variables passed to other functions that expect paths, e.g. `fopen`.

login/login.c
```diff
@@ -144,7 +144,7 @@ int request_url(const uint8_t service_key[GLOME_MAX_PUBLIC_KEY_LENGTH],
                 const uint8_t prefix_tag[GLOME_MAX_TAG_LENGTH],
                 size_t prefix_tag_len, char** url, int* url_len,
                 const char** error_tag) {
-  if (prefix_tag_len >= GLOME_MAX_TAG_LENGTH) {
+  if (prefix_tag_len > GLOME_MAX_TAG_LENGTH) {
```
burgerdev, line 147: Good catch! Could you please add a test to prevent regression?

burgerdev: And maybe consider fixing this in a separate PR so that it does not depend on the naming changes.

login/ui.h
Republicn: This comments are no more actual for this PR. You may find according changes here: https://github.com/google/glome/pull/64.

pkern (approved):

PR #55, opened 2020-09-23
Golang glome-login library

Republicn:
This is a golang implementation of glome-login protocol.


go/login/login.go
```diff
@@ -0,0 +1,306 @@
...
+	response     *URLResponse
+}
+
+// Client constructor. Sets Client.ServiceKey, Client.UserKey, Client.ServiceKeyId to the corresponding values
+// and Client.response to nil.
+func NewClient(sk glome.PublicKey, uk glome.PrivateKey, sId uint8) *Client {
+	return &Client{sk, uk, sId, nil}
+}
+
+func (c *Client) Construct(V byte, hostIdType string, hostId string, action string) (string, error) {
```
burgerdev, line 148: Please document what is constructed. I'd prefer if both the client and the server struct would be operating on the same type (`*url.URL` in this case). However, this would require knowing about the scheme and host, which is also inconvenient.

go/login/login.go
```diff
@@ -0,0 +1,306 @@
...
+	r := c.response
+
+	handshake = append(handshake, r.d.User[:]...)
+	handshake = append(handshake, r.GetTag()[:]...)
+	return base64.URLEncoding.EncodeToString(handshake[:])
+}
+
+// Validates if the received tag corresponding to the base64-url encoded message constructed from the Message.
+func (c *Client) ValidateAuthCode(tag string) bool {
+	return c.response.ValidateAuthCode(tag)
```
burgerdev, line 175: Please make sure that `c.response` is never used uninitialized (which will panic).

go/login/login.go
```diff
@@ -0,0 +1,306 @@
...
+func (s *Server) getResponse() *URLResponse {
+	return s.response
+}
+
+func (s *Server) ParseURLResponse(url *url.URL) (*URLResponse, error) {
+	path := url.Path
+	response := URLResponse{}
+
+	exp := "/(?P<v>v([1-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5]))/(?P<handshake>\\w+)/"
+	r := regexp.MustCompile(exp)
```
burgerdev, line 201: `r` should be a package-level variable (with a different name ^^).

go/login/login.go
```diff
@@ -0,0 +1,306 @@
...
+
+func (s *Server) getResponse() *URLResponse {
+	return s.response
+}
+
+func (s *Server) ParseURLResponse(url *url.URL) (*URLResponse, error) {
+	path := url.Path
+	response := URLResponse{}
+
+	exp := "/(?P<v>v([1-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5]))/(?P<handshake>\\w+)/"
```
burgerdev, line 200: Do we need to verify the `V` range in the regexp? Doing it in code may be easier to read.

burgerdev, line 200: `\w` does not include all url-safe base64 chars.

go/login/login.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+func (c *Client) getResponse() *URLResponse {
+	return c.response
+}
+
+type Server struct {
+	KeyFetcher func(uint8) glome.PrivateKey
+	UserKey    glome.PublicKey
+	response   *URLResponse
```
burgerdev, line 185: Why do we need the `response` on the `Server` struct when it is also returned by `ParseURLResponse`?

go/login/login.go
```diff
@@ -0,0 +1,306 @@
...
+func NewServer(fetcher func(uint8) glome.PrivateKey, k glome.PublicKey) *Server {
+	return &Server{fetcher, k, nil}
+}
+
+func (s *Server) getResponse() *URLResponse {
+	return s.response
+}
+
+func (s *Server) ParseURLResponse(url *url.URL) (*URLResponse, error) {
+	path := url.Path
```
burgerdev, line 197: You will probably need to use `RawPath` so that you can tell the difference between encoded slashes and real ones. https://golang.org/pkg/net/url/#URL

go/glome/glome.go
```diff
@@ -231,6 +231,19 @@ func (d *Dialog) Tag(msg []byte, counter uint8) []byte {
...
 func (d *Dialog) Check(tag []byte, msg []byte, counter uint8) bool {
+	expect := d.expect(tag, msg, counter)
+	return hmac.Equal(expect, tag)
+}
+
+// CheckEncoded method checks if an encoded tag matches some provided message and the counter.
+// The used encoding is base64-url. For more details look at the Check method description.
+func (d *Dialog) CheckEncoded(tag string, msg []byte, counter uint8) bool {
+	expect := d.expect([]byte(tag), msg, counter)
+	return base64.URLEncoding.EncodeToString(expect) == tag
```
pkern, line 242: hmac.Equal?

go/glome/glome.go
```diff
@@ -231,6 +231,19 @@ func (d *Dialog) Tag(msg []byte, counter uint8) []byte {
...
+}
+
+// Returns the expected tag according to the provided message and the counter.
+func (d *Dialog) expect(tag []byte, msg []byte, counter uint8) []byte {
```
pkern, line 246: I think I'd go with `expected` rather than `expect`. Your function does not do an assertion and expects something, it returns an expected result.

go/login/login.go
```diff
@@ -0,0 +1,306 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package login
+
+import (
+	"encoding/base64"
+	"fmt"
+	"github.com/google/glome/go/glome"
```
pkern, line 20: Sort this into a separate import block:

```go
import (
  "stdlib"

  "github.com/google/glome/..."
)
```

go/login/login.go
```diff
@@ -0,0 +1,306 @@
...
+	"net/url"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+const (
+	// Minimal acceptable length of a handshake. 1 byte for the prefix, 32 bytes for the key
+	MinHandshakeLen = 1 + glome.PublicKeySize
```
pkern, line 29: Does this need to be public? Would a caller use this for something?

go/login/login.go
```diff
@@ -0,0 +1,306 @@
...
+)
+
+var (
+	// ErrFailedHandshake denotes that the URL has a wrong format.
+	ErrInvalidURLFormat = fmt.Errorf("url is malformed")
```
pkern, line 34: s/url/URL/

go/login/login.go
```diff
@@ -0,0 +1,306 @@
...
+	// ErrFailedHandshake denotes that the handshake is too short.
+	ErrInvalidHandshakeLen = fmt.Errorf("handshake length is small: should be at least %d", MinHandshakeLen)
+	// ErrVersionNotSupported denotes that the version of glome-login URL format is not supported.
+	ErrVersionNotSupported = fmt.Errorf("version not supported")
+	// ErrInvalidPrefixType denotes that the prefix type is invalid.
+	ErrInvalidPrefixType = fmt.Errorf("invalid prefix type")
```
pkern, line 40: I'm not super happy about errors that do not include what the wrong input was. Sometimes that makes sense and `ErrInvalidHandshakeLen` is totally fine. But "version not supported" should be something like "version not supported: 2". Same at least for "invalid prefix type". If "URL is malformed" could state more clearly what went wrong, that'd also be good.

In general Go code just does `return fmt.Errorf("URL is malformed: %v", err)` (or even `%w` nowadays) and there are not fixed types that are returned. Types are sometimes useful, especially if you want to match in *application* code on the concrete error type. The following satisfies the error interface just fine:

```go
type ErrVersionNotSupported int

func (e ErrVersionNotSupported) Error() string {
  return fmt.Sprintf("version not supported: %d", e)
}
```

(Or the equivalent with a struct.)

go/login/login.go
```diff
@@ -0,0 +1,306 @@
...
+)
+
+// Contains the <message> information
```
pkern, line 43: I think this comment should elaborate what a Message is and what it represents rather than this generic statement.

go/login/login.go
```diff
@@ -0,0 +1,306 @@
...
+type Message struct {
+	HostIdType string
+	HostId     string
+	Action     string
+}
+
+// Constructs a Message according to the structure: [<hostid-type>:]<hostid>[/<action>].
+func (msgInfo *Message) Construct() []byte {
```
pkern, line 51: Receiver names are generally very short. So I'd write `(m *Message)` instead.

pkern (changes requested):

go/glome/glome.go
Republicn: `hmac.Equal()` works only with byte arrays, while here I compare strings. I guess that is the only way to check encoded tags when the received one is truncated and it's bit length (without padding) is not divided by 8 (like in the [first test vector](https://github.com/google/glome/blob/master/docs/glome-login.md#vector-1), where the response tag is truncated to the ten-letter word and is represented with 60bits).

go/login/login.go
Republicn: Fixed

go/login/login.go
Republicn: Fixed.

go/login/login.go
Republicn: Fixed.

go/glome/glome.go
Republicn: Agree, thank you. Fixed.

go/login/login.go
Republicn: Fixed.

Republicn: Must be fixed now.

go/login/login.go
Republicn: Fixed.

go/login/login.go
Republicn: Updated to `string` usage instead of `*url.URL`. Had thought about both variant, but firstly decided the second variant as it contains scheme which is sometimes in the url.

go/glome/glome.go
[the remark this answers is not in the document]
Republicn: Changed: now pass an encoding func as a parameter. WDYT?

go/login/login.go
Republicn: Fixed.

go/login/login.go
Republicn: You are right, thanks. Removed.

go/login/login.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func (c *Client) getResponse() *URLResponse {
+	return c.response
+}
+
+// Server side login-lib handler.
+type Server struct {
+	KeyFetcher func(uint8) glome.PrivateKey
+	UserKey    glome.PublicKey
```
pedrobn23, line 195: Have some questions about this:
- How should the server initialize the UserKey parameter? If it should not be initalized by the user consider making it unexported.
- Can a single Server deals with multiple calls for `ParseURLResponse` done concurrently, or, should a server object be created for each request?

pedrobn23: If not, what should the KeyFetcher reply if no key is found? would it be possible that KeyFercher return an error or a bool as well as a `glome.PrivateKey`?

go/glome/glome.go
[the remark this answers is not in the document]
burgerdev: I'm still opposed to adding this to the glome library. The complications of truncated base64 encoding are glome-login's fault and should thus be dealt with there. Can we just try to reconstruct as much of the tag as possible from the truncated base64 and pass that to `glome.Check`? https://play.golang.org/p/MXuNeI91i_M

go/login/login.go
burgerdev: Makes sense, but please update the documentation because we're not adding a `url-prefix` here.

go/login/login.go
```diff
@@ -0,0 +1,313 @@
...
+	// ErrVersionNotSupported denotes that the version of glome-login URL format is not supported.
+	ErrVersionNotSupported = fmt.Errorf("version not supported")
+	// ErrInvalidPrefixType denotes that the prefix type is invalid.
+	ErrInvalidPrefixType = fmt.Errorf("invalid prefix type")
+	// ErrIncorrectTag denotes that received tag is incorrect.
+	ErrIncorrectTag = fmt.Errorf("invalid tag")
+)
+
+var (
+	validURLPrefix = regexp.MustCompile(`/(?P<v>v[1-9][0-9]*)/(?P<handshake>[\w-_=]+)/`)
```
burgerdev, line 47: Although it seems to be allowed in RE2, I'd prefer the minus sign to come last in the character class (as [specified](https://pubs.opengroup.org/onlinepubs/009696899/basedefs/xbd_chap09.html#tag_09_03_05) for POSIX regular expressions). `\w` seems to include underscore already.

Republicn: Thank you, fixed!

go/login/login.go
Republicn: Removed redundant field.

go/login/login.go
Republicn: Fixed.

go/glome/glome.go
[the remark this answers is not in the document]
Republicn: Thank you for the suggestion, implemented it.

Republicn: Resolving the conversation as there it is not actual anymore.

go/login/login.go
```diff
@@ -0,0 +1,337 @@
...
+	res := url.PathEscape(s)
+	for _, c := range "!*'();:@&=+$,[]" {
+		st := string(c)
+		strings.Replace(res, url.PathEscape(st), st, -1)
+	}
+	return res
+}
+
+// Handshake struct represents the context required for constructing the handshake.
+// It contains Prefix - either service key or it's id, UserKey - user's public ephemeral key,
```
pkern, line 85:
```suggestion
// It contains Prefix - either service key or its id, UserKey - user's public ephemeral key,
```

pkern (changes requested):

go/login/login.go
```diff
@@ -0,0 +1,337 @@
...
+func NewResponse(serviceKeyId uint8, serviceKey glome.PublicKey, userKey glome.PrivateKey,
+	V byte, hostIdType string, hostId string, action string) (*URLResponse, error) {
+	var prefix byte
+	var r URLResponse
+
+	r.V = V
+
+	d, e := userKey.TruncatedExchange(&serviceKey, 1) // TODO: receive m as param?
+	if e != nil {
+		return nil, e
```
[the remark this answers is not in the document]
Republicn: Renamed.

go/login/login.go
Republicn: Added a check for a `nil` value.

go/login/login_test.go
```diff
@@ -0,0 +1,181 @@
...
+	"fmt"
+	"strconv"
+	"testing"
+
+	"github.com/google/glome/go/glome"
+)
+
+const (
+	tvsCnt = 2
+)
```
pkern, line 28: Can't you just use `len()` instead of defining a constant here? `[][]typ` is a valid type (slice of slices).

go/login/login_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+var serviceKeyIds = []uint8{1, 0}
+
+func handleError(e error, t *testing.T) {
+	if e != nil {
+		t.Fatalf("Unexpected Error: " + e.Error())
+	}
+}
```
pkern, line 36: I'm not a fan of error handling functions. Generally error checking is inlined, also because it can include more context. E.g. you can write `t.Fatalf("PrivateKeyFromSlice failed unexpectedly: %v", err)` and whoever reads the test result knows which function has failed. Generally Go does not print a backtrace because these are not exceptions so it's harder to figure out where an error came from.

go/login/login_test.go
```diff
@@ -0,0 +1,181 @@
...
+	hostId     string
+	action     string
+	msg        []byte
+	url        string
+	prefixN    []byte
+	tag        []byte
+	token      string
+}
+
+type KeyPair struct {
```
pkern, line 55: Unexport local test-only type

go/login/login_test.go
```diff
@@ -0,0 +1,181 @@
...
+	handleError(err, t)
+	receiving, err := serverKP.priv.Exchange(&clientKP.pub)
+	handleError(err, t)
+
+	return sending, receiving
+}
+
+func testVectors() []testVector {
+	prefix1, _ := strconv.Atoi("1")
+	prefix2, _ := strconv.Atoi("0x51")
```
pkern, line 90: Never swallow errors. In this case I'm fairly sure 0x51 is not a legit number? godoc says "Atoi is equivalent to ParseInt(s, 10, 0), converted to type int." and you are looking for base 16. And indeed https://play.golang.org/p/hOHLM8F9rVJ throws an error.

go/login/login_test.go
```diff
@@ -0,0 +1,181 @@
...
+			msg:        []byte("serial-number:1234567890=ABCDFGH/#?/reboot"),
+			url:        "/v1/UYcvQ1u4uJ0OOtYqouURB07hleHDnvaogAFBi-ZW48N2/serial-number:1234567890=ABCDFGH%2F%23%3F/reboot/",
+			prefixN:    decodeString("dff5aae753a8bdce06038a20adcdb26c7be19cb6bd05a7850fae542f4af29720"),
+			tag:        decodeString("06476f1f314b06c7f96e5dc62b2308268cbdb6140aefeeb55940731863032277"),
+			token:      "p8M_BUKj7zXBVM2JlQhNYFxs4J-DzxRAps83ZaNDquY=",
+		},
+	}
+}
+
+func ClientsAndServers(t *testing.T) ([]Client, []Server) {
```
pkern, line 129: unexport local function

go/login/login_test.go
```diff
@@ -0,0 +1,181 @@
...
+			for _, k := range []struct {
+				expected string
+				got      string
+			}{
+				{expected: tv.hostIdType, got: responses[i].Msg.HostIDType},
+				{expected: tv.hostId, got: responses[i].Msg.HostID},
+				{expected: tv.action, got: responses[i].Msg.Action},
+			} {
+				if k.expected != k.got {
+					t.Fatalf("TestCheckCorrectURLs failed for test vector %d. Expected: %v, got: %v", i, k.expected, k.got)
```
pkern, line 176: How does the error actually look? `%#+v` might potentially be a better string format for expected and got.

pkern (changes requested):

go/login/login_test.go
Republicn: Fixed for testing. Need to have a look at login.go

go/login/login_test.go
Republicn: Fixed.

go/login/login_test.go
Republicn: Removed.

go/login/login_test.go
Republicn: Fixed

go/login/login_test.go
Republicn: Fixed

go/login/login.go
```diff
@@ -0,0 +1,367 @@
+// Copyright 2020 Google LLC
```
pedrobn23, line 1: May I suggest running golint?

Republicn: Thank you for the suggestion! Done.

go/login/login.go
Republicn: Fixed.

go/login/login.go
Republicn: Not actual anymore.

go/glome/glome.go
```diff
@@ -16,11 +16,10 @@
 package glome
```
pedrobn23, line 16: Nit: imho small changes in code that does not alter the functionality of the code should be avoided so that every commit is self-contained.

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+	return res
+}
+
+// Handshake struct represents the context required for constructing the handshake.
+// It contains Prefix - either service key or its id, UserKey - user's public ephemeral key,
+// MsgTagPrefix - tag calculated under Message.
+type Handshake struct {
+	Prefix       byte
+	UserKey      glome.PublicKey
+	MsgTagPrefix []byte
```
pedrobn23, line 104: nit: should abreviation be Msg or MSG?

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+	case 2:
+		return s + "=="
+	case 3:
+		return s + "="
+	default:
+		panic("math fail")
+	}
+}
+
+func (c *Client) getResponse() *URLResponse {
```
pedrobn23, line 247: https://golang.org/doc/effective_go.html#Getters

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+	return c.response
+}
+
+// Server side glome-login lib handler. Receives the server's private key fetcher function,
+// which returns an error if the key couldn't be calculated.
+type Server struct {
+	KeyFetcher func(uint8) (glome.PrivateKey, error)
+}
+
+// ParseURLResponse parses the url, checks whether it is formed correctly and validates the client's tag, received from the URL.
```
pedrobn23, line 257: Add list of errors to documentation.

go/login/login_test.go
```diff
@@ -0,0 +1,224 @@
...
+	clientTagsLen := []uint{2, 0}
+	keyPairs := make([][]keyPair, len(tvs))
+	for i, tv := range tvs {
+		keyPairs[i] = append(keyPairs[i], *keys(t, tv.kap, tv.ka), *keys(t, tv.kbp, tv.kb))
+	}
+
+	var clients []Client
+	var servers []Server
+	for tv := 0; tv < len(tvs); tv++ {
+		clients = append(clients, *NewClient(keyPairs[tv][1].pub, keyPairs[tv][0].priv, serviceKeyIDs[tv], clientTagsLen[tv]))
```
pedrobn23, line 138: Why use `KeyPairs[tv][1]` instead of `tv.kbp`?

pedrobn23 (approved):
LGTM.

go/login/login.go
pedrobn23: Also apply to getTag

go/login/login.go
burgerdev: I believe it should be Msg (i.e. only capitalized if it's an actual acronym). However, may I suggest calling it `MessageTagPrefix`?

go/glome/glome.go
Republicn: SGTM too.

go/login/login.go
Republicn: Fixed.

go/login/login.go
Republicn: @pedrobn23 I still think it shouldn't be capitalized.
@burgerdev sgtm, fixed.

go/login/login_test.go
```diff
@@ -0,0 +1,235 @@
...
+func TestURLResponseConstruction(t *testing.T) {
+	clients, _ := clientsAndServers(t)
+	for i, tv := range tvs {
+		t.Run("Test vector "+fmt.Sprint(i+1), func(t *testing.T) {
+			resp, err := clients[i].Construct(1, tv.hostIDType, tv.hostID, tv.action)
+			if err != nil {
+				fatal(fmt.Sprintf("Error while constructing URL: %s.", err.Error()), t, "TestURLResponseConstruction", i+1)
+			}
+
+			if !(resp == tv.url) {
```
burgerdev, line 221: Consider `!=` instead.

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+
+	if m.Action != "" {
+		action = "/" + m.Action
+	}
+	return []byte(hostIDType + hostID + action)
+}
+
+// Escapes the string so it can be safely placed inside a URL path segment,
+// replacing "/#?" special characters and not replacing "!*'();:@&=+$,[]" special characters.
```
pkern, line 88: Filed https://github.com/google/glome/issues/62 so that we eventually address the imprecise wording in the documentation. I feel like we had more bits that needed fixing, but alas.

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+	for _, c := range "!*'();:@&=+$,[]" {
+		st := string(c)
+		strings.Replace(res, url.PathEscape(st), st, -1)
+	}
+	return res
+}
+
+// Handshake struct represents the context required for constructing the handshake.
+// It contains Prefix - either service key or its id, UserKey - user's public ephemeral key,
+// MessageTagPrefix - tag calculated under Message.
```
pkern, line 100: Same comments as `Message` above.

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+type Handshake struct {
+	Prefix           byte
+	UserKey          glome.PublicKey
+	MessageTagPrefix []byte
+}
+
+// URLResponse struct represents the context required for the URL constructing.
```
pkern, line 107:
```suggestion
// URLResponse represents the context required for the construction of the URL.
```

Also same: move struct member comments to the individual members.

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+	d, err := userKey.TruncatedExchange(&serviceKey, 1)
+	if err != nil {
+		return nil, err
+	}
+	r.d = d
+
+	r.Msg = Message{hostIDType, hostID, action}
+
+	if serviceKeyID == 0 {
+		// If no key ID was specified, send the first key byte as the ID.
```
pkern, line 134:
```suggestion
		// If no key ID was specified, send the first key byte as the ID.
		// TODO(#60): Fix this up once there is clarify on key prefix usage.
```

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+		prefix = serviceKey[0] & 0x7f
+	} else {
+		prefix = serviceKeyID & 0x7f
+	}
+	userPublic, err := userKey.Public()
+	if err != nil {
+		return nil, err
+	}
+	HandshakeInfo := Handshake{prefix, *userPublic, r.Tag(tagLen)}
+	r.HandshakeInfo = HandshakeInfo
```
pkern, line 144: Fold into one line.

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+
+	return &r, nil
+}
+
+// ValidateAuthCode checks if the received tag corresponding to the base64-url encoded message constructed from the Message.
+// Returns true if the received tag is empty.
```
pkern, line 150: I'm confused about the comment. What is it that the function is doing? When would you call it? Why is an empty tag true?

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+func (r *URLResponse) ValidateAuthCode(tag []byte) bool {
+	if len(tag) == 0 {
+		return true
+	}
+	return r.d.Check(tag, r.Msg.Construct(false), 0)
+}
+
+// Tag returns the tag corresponding to the Msg. The returned tag is calculated with usage of sendingKey.
```
pkern, line 158: What is sendingKey? Is that an older name that did not survive refactoring?

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+func (r *URLResponse) Tag(len uint) []byte {
+	return r.d.Tag(r.Msg.Construct(false), 0)[:len]
+}
+
+// EncToken returns a base64-encoded response token.
+func (r *URLResponse) EncToken() string {
+	return base64.URLEncoding.EncodeToString(r.Tag(glome.MaxTagSize)) // TODO: passing the tag len as param?
+}
+
+// Client side glome-login handler. Should be constructed under NewClient constructor.
```
pkern, line 168: Consider writing prose. "Client implements the client-side of the glome-login protocol." maybe. Given that there is an unexported field it is clear that you must use a constructor and in the public API docs I'd just omit what that is. It should be clear from the documentation. That being said, would it have been possible not to have an unexported field? The constructor does not seem to assert anything. Could we just generate response where we need it, does it make sense to retain it at all given that it likely changes with every message anyway?

pkern (changes requested):

go/login/login.go
Republicn: Will do!

go/login/login.go
Republicn: Fixed.

go/login/login_test.go
Republicn: Fixed.

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+	if len(msgTagPrefix) > glome.MaxTagSize {
+		return nil, glome.ErrInvalidTagSize
+	}
+
+	return &Handshake{prefix, *userKey, msgTagPrefix}, nil
+}
+
+// parseMsg returns the parsed version of the URL message.
+// The message should satisfy the following format: [<hostid-type>:]<hostid>[/<action>].
+func parseMsg(m []string) (*Message, error) {
```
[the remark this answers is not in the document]
Republicn: Thank you for the suggestion. Make it receive a string to parse.

go/login/login.go
Republicn: No, it's a key calculated by a `Dialog` for a sending party.

go/login/login.go
Republicn: Added annotations.

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+	ErrServerKeyNotFound = fmt.Errorf("server key not found")
+)
+
+var (
+	validURLPrefix = regexp.MustCompile(`/(?P<v>v[1-9][0-9]*)/(?P<handshake>[\w=-]+)/`)
+)
+
+// Message struct represents the context required for authorization.
+// It contains HostIDType - type of identity, HostID - identity of the target (e.g. hostname, serial number, etc.),
+// Action - action that is being authorized.
```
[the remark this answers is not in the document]
Republicn: Added.

go/login/login_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+type keyPair struct {
+	priv glome.PrivateKey
+	pub  glome.PublicKey
+}
+
+func decodeString(s string) []byte {
+	b, err := hex.DecodeString(s)
+	if err != nil {
+		panic(fmt.Sprintf("Invalid hexadecimal string %v input in test", s))
```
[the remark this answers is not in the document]
Republicn: Fixed for `t.Fatalf` for now.

go/login/login.go
Republicn: Thank you, some of the errors became structs.

go/login/login.go
Republicn: Added.

go/login/login.go
```diff
@@ -0,0 +1,388 @@
...
+	ErrVersionNotSupported = fmt.Errorf("version not supported")
+	// ErrInvalidPrefixType denotes that the prefix type is invalid.
+	ErrInvalidPrefixType = fmt.Errorf("invalid prefix type")
+	// ErrIncorrectTag denotes that received tag is incorrect.
+	ErrIncorrectTag = fmt.Errorf("invalid tag")
+	// ErrResponseNotInitialized denotes that the response is not initialized.
+	ErrResponseNotInitialized = fmt.Errorf("response is not initialized")
+	// ErrServerKeyNotFound server key not found.
+	ErrServerKeyNotFound = fmt.Errorf("server key not found")
+)
```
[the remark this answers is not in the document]
Republicn: Fixed.

go/login/login.go
Republicn: Updated. The empty tag is ok for client's tag. Moved this logic to ParseURLResponse.

go/login/login.go
Republicn: Updated. What do you mean by "It should be clear from the documentation."? I've mentioned it in the docstring.
About the URLResponse - it is used later for ValidateAuthCode, which should be called on Client rather than on URLResponse (which needs to be done for Server). I guess that is an imperfection of an API and ValidateAuthCode should be also implemented on Server and responses method should become private. Don't know if it worse changing at the moment.

PR #56, opened 2020-09-23
Add python fuzzing test

pedrobn23:
Add file python/test/fuzzing_test.py that implements some simple fuzz testing. Also modify README to document this.

Signed-off-by: pedrobn <pedrobn@google.com>

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -76,7 +76,9 @@ from this directory. If you only want to execute a particular test module, then 
 python -m test.my_module_name
 ```
 
-where `my_module_name` is the name of the test module to be executed (the name of the file without the .py). A simple fuzzing test was also done and can be found on `fuzzing_test.py`. This test is not run along the unittest, to run it use:
+where `my_module_name` is the name of the test module to be executed (the name of the file without the .py). 
+
+To run the fuzzing test use:
 ```
 python -m test.fuzzing_test
 ```
--- a/python/README.md
+++ b/python/README.md
@@ -65,7 +65,7 @@ For more information see the in-code documentation.
 
 ### Test
 
-In the test folder we have scripts that implement test classes based on unittest. To run all tests use:
+In the test folder we have scripts that implement test classes based on unittest. To run all unittest use:
 
 ```
 python -m test
@@ -76,4 +76,8 @@ from this directory. If you only want to execute a particular test module, then 
 python -m test.my_module_name
 ```
 
-where `my_module_name` is the name of the test module to be executed (the name of the file without the .py).
+where `my_module_name` is the name of the test module to be executed (the name of the file without the .py). A simple fuzzing test was also done and can be found on `fuzzing_test.py`. This test is not run along the unittest, to run it use:
+```
+python -m test.fuzzing_test
+```
+from this directory.
--- a/python/test/fuzzing_test.py
+++ b/python/test/fuzzing_test.py
@@ -129,6 +129,7 @@ def autoglome_test(private_bytes, public_bytes, msg, tag, counter, min_tag_len,
     _autoglome(private_bytes, public_bytes, msg, tag, counter, min_tag_len,
                skippable)
 
+
 @hypothesis.settings(max_examples=10**5)
 @hypothesis.given(
     hypothesis.strategies.binary(),  #private_bytes
@@ -137,7 +138,8 @@ def autoglome_test(private_bytes, public_bytes, msg, tag, counter, min_tag_len,
     hypothesis.strategies.binary(min_size=32, max_size=32),  #tag
     hypothesis.strategies.integers(),  #counter
     hypothesis.strategies.integers())  #min_tag_len
-def glome_unsized_keys_test(private_bytes, public_bytes, msg, tag, counter, min_tag_len):
+def glome_unsized_keys_test(private_bytes, public_bytes, msg, tag, counter,
+                            min_tag_len):
     """Add hypothesis decorator to _glome function"""
     _glome(private_bytes, public_bytes, msg, tag, counter, min_tag_len)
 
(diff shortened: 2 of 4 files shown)


python/test/fuzzing_test.py
```diff
@@ -0,0 +1,122 @@
...
+    except (ValueError, pyglome.ExchangeError):
+        return 
+
+    try:
+        tag_manager = pyglome.Glome(peer_key, private_key, min_tag_len)
+    except ValueError:
+        return
+
+    tag_manager.user_keys
+    tag_manager.peer_key
```
pkern, line 39: Are you doing this for the side effects? If so, please add a comment.

python/test/fuzzing_test.py
```diff
@@ -0,0 +1,122 @@
...
+
+    try:
+        tag_manager = pyglome.AutoGlome(peer_key, private_key,
+                                    min_peer_tag_len=min_tag_len,
+                                    skippable_range=skippable_range)
+    except ValueError:
+        return
+
+    tag_manager.sending_counter
+    tag_manager.receiving_counter
```
pkern, line 70: Same here.

python/test/fuzzing_test.py
```diff
@@ -0,0 +1,122 @@
...
+    
+@settings(max_examples=10**7)
+@given(strategies.binary(min_size=32, max_size=32),            #private_bytes
+                  strategies.binary(min_size=32, max_size=32), #public_bytes
+                  strategies.binary(),                         #msg
+                  strategies.binary(min_size=32, max_size=32), #tag
+                  strategies.integers(),                       #counter
+                  strategies.integers())                       #min_tag_len
+def glome_test(private_bytes, public_bytes, msg, tag, counter, min_tag_len):
+    """Add hypothesys decorator to _glome function"""
```
pkern, line 105: "hypothesis"?

python/test/fuzzing_test.py
```diff
@@ -0,0 +1,122 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Python GLOME fuzz main.
+
+This module that implement some easy fuzz testing for pyglome. 
+"""
+
+from hypothesis import given, settings, strategies, example
```
pkern, line 19: pystyle rule 2.2 "Use import statements for packages and modules only, not for individual classes or functions."

python/README.md
```diff
@@ -76,4 +76,8 @@ from this directory. If you only want to execute a particular test module, then
 python -m test.my_module_name
 ```
 
-where `my_module_name` is the name of the test module to be executed (the name of the file without the .py).
+where `my_module_name` is the name of the test module to be executed (the name of the file without the .py). A simple fuzzing test was also done and can be found on `fuzzing_test.py`. This test is not run along the unittest, to run it use:
+```
```
pkern, line 80: CI should run tests that exist to ensure they do not break. Even if it doesn't do true fuzzing and only runs one iteration.

python/test/fuzzing_test.py
```diff
@@ -0,0 +1,122 @@
...
+from cryptography.hazmat.primitives.asymmetric import x25519
+
+import pyglome
+
+
+def _glome(private_bytes, public_bytes, msg, tag, counter, min_tag_len):
```
pkern, line 25: So the intention here is to find cases that throw unexpected exceptions?

python/README.md
pedrobn23: SGTM! A "not real fuzzing" call to both function was added in a unittest class and included in the call to `python -m test` . I'm thinking about not changing the readme in order not to mislead about whether `python -m test` does do fuzzing, wdyt?

python/test/fuzzing_test.py
pedrobn23: Fixed

python/test/fuzzing_test.py
pedrobn23: Yes, it is. Added in the function docstring.

python/test/fuzzing_test.py
pedrobn23: Added

python/test/fuzzing_test.py
pedrobn23: Fixed

python/test/fuzzing_test.py
```diff
@@ -0,0 +1,169 @@
...
+    try:
+        tag_manager.check(tag, msg)
+    except (pyglome.IncorrectTagError, pyglome.TagCheckError):
+        pass
+
+
+@hypothesis.settings(max_examples=10**7)
+@hypothesis.given(
+    hypothesis.strategies.binary(min_size=32, max_size=32),  #private_bytes
+    hypothesis.strategies.binary(min_size=32, max_size=32),  #public_bytes
```
burgerdev, line 107: Have you considered generating incorrect key sizes as well? These should also throw only documented exceptions, but they should probably be fuzzed separately so that we still get coverage on the tagging.

python/README.md
```diff
@@ -65,7 +65,7 @@ For more information see the in-code documentation.
 
 ### Test
 
-In the test folder we have scripts that implement test classes based on unittest. To run all tests use:
+In the test folder we have scripts that implement test classes based on unittest. To run all unittest use:
```
burgerdev, line 69: Can you please add documentation on how to run the fuzz tests?

python/test/fuzzing_test.py
pedrobn23: Test Added!

python/README.md
pedrobn23: Information added on next parragraph

burgerdev (approved):

PR #70, opened 2020-12-14
Fix program name used for syslog messages

vvidic:


l9i (approved):
Ooops. It's an leftover from the previous Google-internal version of GLOME.

Thanks for fixing this.

Issue #39, opened 2020-08-07
Provide pam_glome.so (PAM module for GLOME)


l9i:
Implementing a [PAM module](https://en.wikipedia.org/wiki/Pluggable_authentication_module) for GLOME would allow for an easier adoption of GLOME.

pkern:
I think for this we would want to reuse glome-login's configuration and URL generation infrastructure - so maybe glome-login should actually export some kind of library, although I'd argue that a static one is probably better than a shared one.

One thing to test here is conversation support. We would need to display our own prompt and reuse the AUTHTOK as the response to the challenge. This should be possible but relies on the PAM-using application actually being able to display a prompt.

pkern:
I put a skeleton module into [pkern/glome@pam](/pkern/glome/tree/pam).

pkern:
Thanks to Valentin we have a module now! \o/

The other part that I think is still missing is some sort of testing regime. I'm not sure how easy that is with a challenge-response module. [pam_wrapper](https://lwn.net/Articles/671094/) and [fakepam](https://www.eyrie.org/~eagle/software/rra-c-util/fakepam.html) come to mind for testing.

vvidic:
Ah yes, I'll try to add a `pam_test` binary for this.

vvidic:
I suppose we can close this one now the tests are also merged? :)

Issue #33, opened 2020-08-05
glome-login refers to a mysterious "embedded" service key


delroth:
In `glome-login --help` I spotted the following:

```
 -k KEY    use KEY as the hex-encoded service key (defaults to the embedded key)
```

I can't seem to find any reference to what this embedded key is and where it lives, and from looking at the source code I suspect it might not exist. Could you clarify the documentation and/or remove this?

bluecmd:
Internally we do not use a config file, we instead call `login_run` straight away with hardcoded public keys - that's where the awkward description comes from.

Right now -k is parsed before config is read, so the config file takes precedence - which is a bit odd.

I'd suggest we:

- Change the description of -k to say `.. (defaults to key from configuration)`
- Change that -k overrides the key possibly present in the configuration file

burgerdev:
The last occurrence I knew of has been removed in #73.

PR #73, opened 2020-12-20
Implement PAM authentication module for GLOME

vvidic:
Reuses much of the existing glome-login functions.

### DIFF
--- a/login/README.md
+++ b/login/README.md
@@ -68,3 +68,41 @@ glome-login uses error tags to communicate errors.
 This error means that `glome-login` could not figure out what service key to
 use. This most likely means that you have not specified a service key in the
 configuration file (by default `/etc/glome/config`).
+
+# PAM module
+
+`pam_glome.so` library implements the PAM authentication module for the
+[GLOME Login](../docs/glome-login.md) protocol.
+
+## Installation
+
+1. Install the library into the system dependent location for PAM modules
+   (for example `/lib/security/pam_glome.so`).
+1. Enable and configure PAM module for a specific service (for example
+   `/etc/pam.d/login`):
+
+```
+auth       requisite  pam_glome.so
+```
+
+## Usage
+
+PAM module supports the following options:
+
+* `config_path=PATH` - location of the configuration file to parse (defaults to
+  `/etc/glome/config`)
+* `service_key=KEY` - use hex-encoded `KEY` as the service key (defaults to key
+  from configuration file)
+* `service_key_version=N` - use `N` for the service key version (defaults to key
+  version from configuration file)
+* `url_prefix=URL` - use given URL prefix (defaults to prefix from configuration
+  file)
+* `debug` - enable verbose logging
+* `insecure_debug` - enable logging of secrets (INSECURE!)
+* `insecure_host_id=NAME` - use `NAME` as the host-id
+* `insecure_secret_key=KEY` - use hex-encoded `KEY` instead of the ephemeral
+  secret key (INSECURE!)
+
+## Troubleshooting
+
+PAM module uses error tags to communicate errors in the syslog messages.
--- a/login/meson.build
+++ b/login/meson.build
@@ -23,6 +23,8 @@ login_lib = shared_library(
     'glome-login',
     [
         'base64.h',
+        'config.c',
+        'config.h',
         'crypto.c',
         'crypto.h',
         'lockdown.c',
@@ -32,7 +34,7 @@ login_lib = shared_library(
         'ui.c',
         'ui.h',
     ],
-    dependencies : [openssl_dep, glib_dep],
+    dependencies : [openssl_dep, glib_dep, confuse_dep],
     link_with : glome_lib,
     include_directories : glome_incdir,
     install : true, soversion : '0')
@@ -44,14 +46,22 @@ glome_login = executable(
     'glome-login',
     [
         'main.c',
-        'config.c',
-        'config.h',
     ],
-    dependencies : confuse_dep,
     link_with : login_lib,
     include_directories : glome_incdir,
     install : true,
     install_dir : get_option('sbindir'))
+
+cc = meson.get_compiler('c')
+libpam = cc.find_library('pam')
+pam_glome = shared_library('pam_glome',
+    'pam.c',
+    dependencies : [libpam, openssl_dep],
+    link_with : [glome_lib, login_lib],
+    include_directories : glome_incdir,
+    name_prefix : '',
+    install : true,
+    install_dir : join_paths(get_option('libdir'), 'security'))
 
 # TODO: Make tests compilation based on get_option('tests')
 login_test = executable(
--- a/login/pam.c
+++ b/login/pam.c
@@ -1,0 +1,244 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <errno.h>
+#include <limits.h>
+#include <openssl/crypto.h>
+#include <security/pam_ext.h>
+#include <security/pam_modules.h>
+#include <stdlib.h>
+#include <string.h>
+#include <syslog.h>
+#include <time.h>
+
+#include "base64.h"
+#include "config.h"
+#include "lockdown.h"
+#include "login.h"
+#include "ui.h"
+
+#define MODULE_NAME "pam_glome"
+
+static int parse_pam_args(pam_handle_t *pamh, int argc, const char **argv,
+                          login_config_t *config) {
+  memset(config, 0, sizeof(login_config_t));
+  int errors = 0;
+
+  for (int i = 0; i < argc; ++i) {
+    if (!strncmp(argv[i], "config_path=", 12)) {
+      config->config_path = argv[i] + 12;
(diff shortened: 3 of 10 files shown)


vvidic:
Fixes #39

login/pam.c
```diff
@@ -0,0 +1,272 @@
...
+  long l;
+
+  // Setting defaults.
+  config->reboot_user = DEFAULT_REBOOT_USER;
+  config->login_path = DEFAULT_LOGIN_PATH;
+  config->lockdown_path = NULL;
+  config->url_prefix = NULL;
+  config->auth_delay_sec = DEFAULT_AUTH_DELAY;
+  config->input_timeout_sec = DEFAULT_INPUT_TIMEOUT;
+  config->options = SYSLOG;
```
burgerdev, line 47: Some of these options are imho not relevant for the PAM module:

- `auth_delay_sec` should probably be implemented with `pam_faildelay`
- Not entirely sure about `input_timeout_sec`, but I did not find other PAM modules with a configurable timeout.
- Lockdown path is a runtime environment detail that should be handled outside of the GLOME module (c.f. `pam_time`).
- The PAM module arguably should not cause reboots, so we don't need to know about a `reboot_user`.
- The same goes for `login_path`.

Actually, I think it might be better not to use the configuration facilities of `/sbin/glome-login` at all and rather implement only the strictly necessary options (service key, debug, machineid?) as PAM module arguments.

login/pam.c
```diff
@@ -0,0 +1,272 @@
...
+    pam_syslog(pamh, LOG_ERR, "failed to parse pam module arguments (%d)", r);
+    return rc;
+  }
+
+  r = parse_config_file(&config);
+  if (r < 0) {
+    pam_syslog(pamh, LOG_ERR, "failed to read config file: %s (%d)",
+               config.config_path, r);
+    return rc;
+  }
```
burgerdev, line 242: If I have a pam module config that specifies a custom service key and there happens to be a global config file with another service key, the global one would be chosen, wouldn't it? This may be unexpected.

burgerdev (commented):
Thanks a lot for your PR, it's much appreciated!

login/pam.c
```diff
@@ -0,0 +1,272 @@
...
+
+  free(host_id);
+  host_id = NULL;
+  free(action);
+  action = NULL;
+
+  char *input = NULL;
+  if (pam_prompt(pamh, PAM_PROMPT_ECHO_ON, &input,
+                 "%s%s\nObtain GLOME code from the above link: ",
```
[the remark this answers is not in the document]
vvidic: Done.

login/pam.c
```diff
@@ -0,0 +1,272 @@
...
+#include "base64.h"
+#include "config.h"
+#include "lockdown.h"
+#include "login.h"
+#include "ui.h"
+
+#define MODULE_NAME "pam_glome"
+
+static int parse_pam_args(pam_handle_t *pamh, int argc, const char **argv,
+                          login_config_t *config) {
```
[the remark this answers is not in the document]
vvidic: Added module parameter description to README.md.

login/pam.c
```diff
@@ -0,0 +1,272 @@
...
+  } else {
+    pam_syslog(pamh, LOG_ERR, "failed to authenticate user '%s': %s (%d)",
+               config.username, error_tag, r);
+  }
+
+  return rc;
+}
+
+int pam_sm_setcred(pam_handle_t *pamh, int flags, int argc, const char **argv) {
+  return PAM_SUCCESS;
```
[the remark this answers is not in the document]
vvidic: Added a comment for `PAM_SUCCESS` in the new version of the patch.

login/pam.c
vvidic: Yes, this behaviour is mentioned already in #33. Patch is now updated to prefer the parameter over the config file.

login/pam.c
vvidic: Removed the parameters that are not relevant for PAM module.

As for `input_timeout_sec` it seems that PAM prompt waits 60 seconds for input so this should work by default too.

login/pam.c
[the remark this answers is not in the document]
burgerdev: But if we return `PAM_SUCCESS` here, applications will see `pam_setcred(3)` return with `PAM_SUCCESS` as well, won't they? According to the manual, this return value should be interpreted as _Data was successful stored_, and since we're not storing anything may be disingenuous.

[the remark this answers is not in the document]
burgerdev: `PAM_CRED_UNAVAIL: This module cannot retrieve the user's credentials.` sounds correct to me.

vvidic:
On Thu, Jan 07, 2021 at 06:50:51AM -0800, Markus Rudy wrote:
> `PAM_CRED_UNAVAIL: This module cannot retrieve the user's credentials.` sounds correct to me.

I will test with PAM_CRED_UNAVAIL, but I suspect it will cause PAM stack
to fail as mentioned here under RETURN VALUES:

https://man7.org/linux/man-pages/man3/pam_sm_setcred.3.html

Also Google Authenticator uses PAM_SUCCESS for this function:

https://github.com/google/google-authenticator-libpam/blob/master/src/pam_google_authenticator.c#L2192

-- 
Valentin

burgerdev:
> On Thu, Jan 07, 2021 at 06:50:51AM -0800, Markus Rudy wrote: `PAM_CRED_UNAVAIL: This module cannot retrieve the user's credentials.` sounds correct to me.
> I will test with PAM_CRED_UNAVAIL, but I suspect it will cause PAM stack to fail as mentioned here under RETURN VALUES: https://man7.org/linux/man-pages/man3/pam_sm_setcred.3.html Also Google Authenticator uses PAM_SUCCESS for this function: https://github.com/google/google-authenticator-libpam/blob/master/src/pam_google_authenticator.c#L2192
> […](#)
> -- Valentin

I see: if we were to stack GLOME and some password-based login, we would not be able to change the password if the GLOME stack returned UNAVAIL. I don't quite understand the design decision (by PAM), but in that case, and since we have a precedent in Google Authenticator, let's keep it as is. Thanks for following up.

Time to resolve that merge conflict :)

burgerdev (approved):

login/meson.build
```diff
@@ -44,15 +46,21 @@ glome_login = executable(
...
     ],
-    dependencies : confuse_dep,
     link_with : login_lib,
     include_directories : glome_incdir,
     install : true,
     install_dir : get_option('sbindir'))
 
+cc = meson.get_compiler('c')
+libpam = cc.find_library('pam')
+pam_glome = shared_library('pam_glome',
```
pkern, line 57: I was a little tempted to say that this should be conditional. On the other hand I'm of the belief that in the public build glome-login should eventually die and only ship the PAM module. :)

login/pam.c
```diff
@@ -0,0 +1,229 @@
...
+  if (bytes_read < MIN_ENCODED_AUTHCODE_LEN) {
+    free(input);
+    return failure(EXITCODE_INVALID_INPUT_SIZE, error_tag, "authcode-length");
+  }
+  if (bytes_read > strlen(authcode_encoded)) {
+    free(input);
+    return failure(EXITCODE_INVALID_INPUT_SIZE, error_tag, "authcode-length");
+  }
+
+  // We can use insecure strncmp() because it's a *one-time* code.
```
pkern, line 171: Something like this always makes me a little uncomfortable. Is that because we don't have a constant-time check available? You already need to point out that it's safe, so if there's no drawback to just doing the right thing, it'd be better to do that? (Or maybe I'm off here and it just instills a wrong sense of security.)

login/README.md
```diff
@@ -68,3 +68,39 @@ glome-login uses error tags to communicate errors.
...
+auth       requisite  pam_glome.so
+```
+
+## Usage
+
+PAM module supports the following options:
+
+* `config_path=PATH` - location of the configuration file to parse (defaults to
+  `/etc/glome/config`)
+* `service_key=KEY` - use hex-encoded `KEY` as the service key (defaults to key
```
pkern, line 94: Don't we also need a key ID in this case?

login/meson.build
```diff
@@ -44,15 +46,21 @@ glome_login = executable(
...
+    'pam.c',
+    dependencies : [libpam],
+    link_with : [glome_lib, login_lib],
+    include_directories : glome_incdir,
+    install : true)
```
pkern, line 62: Please add a TODO to install this into the correct directory or add the support for that. I know it's messy, at least last time I looked at how to solve this there was no pre-built support in meson.

pkern (commented):
Thank you so much for your contribution! This LGTM as-is except one nit (install dir).

pkern:
And this PR needs to be merged/rebased first to resolve the conflict.

login/pam.c
vvidic: Yes, this is a strange comment and I don't know why is strncmp insecure here since the length is checked few lines before. The comment comes from the original login.c so we should probably check with the author there.

l9i: The original author didn't mean memory safety. :)

It's about `strncmp()` being susceptible to timing attacks which doesn't matter in this particular context given that we are comparing one time code. I am fine with using constant-time comparison here, especially if it makes more people sleep better at nights.

vvidic: Added a XOR loop now, please check if this is ok now.

login/meson.build
vvidic: Added support to install as `$libdir/security/pam_glome.so`

vvidic:
> And this PR needs to be merged/rebased first to resolve the conflict.

PR rebased on top of master.

login/meson.build
vvidic: I suppose there is no harm in building both until the decision is made on glome-login.

login/README.md
vvidic: Added a new option `service_key_version` for setting the key ID.

login/pam.c
burgerdev: Since we already have the OpenSSL dependency I'd prefer using `CRYPTO_memcmp`.

vvidic: Used `CRYPTO_memcmp`, thanks.

PR #78, opened 2021-01-12
Increase portability of GLOME build

### LINKED ISSUE #77: Increase portability of GLOME build
The build process contains a couple of minor choices that prevent porting onto non-glibc systems. My use case is Alpine, but the situation is probably similar on MacOS or *BSD.

- Bash shell is assumed to be available, thus bashisms in scripts.
- Non-POSIX utilities used: `mktemp`,  `realpath`.
- glibc-only headers are used (unnecessarily :) ).

burgerdev:
fixes #77

### DIFF
--- a/cli/test.sh
+++ b/cli/test.sh
@@ -40,7 +40,7 @@ done
 
 errors=0
 for counter in $(seq 0 "$iterations"); do
-  msg="$(head -c 2 /dev/urandom | od -t u2 -A n)"
+  msg="Hello, world!"
   for side in 0 1; do
     peer=$((1 - side))
     tag=$(printf %s "$msg" | "$binary" tag --key "${t}/${side}" --peer "${t}/${peer}.pub" --counter "$counter")
--- a/cli/test.sh
+++ b/cli/test.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 # Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


pkern:
Is getting rid of bash really required? I can understand how it's bad to have a runtime dependency on bash, but for a build-time unit test? I don't think it's useful for us to constrain ourselves to some random superset of POSIX (whatever the shell in Alpine is; for instance Debian mandates certain features to be present in /bin/sh that are not in POSIX). For bash the feature set is known.

burgerdev:
I tried to be careful not to change one superset of POSIX (bash) for another, arbitrary one, but rely on specified features only. The only thing I know may not be fully portable is arithmetic expansion, which is a POSIX extension (not required?) but happens to be implemented virtually everywhere (ksh, mksh, ash, dash, zsh - notable exception is csh). 

There are the usual arguments[1] for having a portable shell (bash-isms may not be portable between bash versions; bash is not available on all systems), but in addition to that I don't see us losing much by restricting to POSIX. The one notable thing is `$RANDOM`, but imho we should even consider throwing that out and using a fixed message.

Admittedly, a runtime dependency would be much worse and we could just specify bash being a build dependency (if tests are desired). 

[1]: https://stackoverflow.com/a/19429855/6144653

pkern:
While you may have been careful now, the next one needs to stick to the same contract and the shell as provided by alpine (and by extension dash, the other default non-bash shell) is going to be the reference. Or is there a static checker that could provide us some assertion that a script might be POSIXly compliant?

vvidic:
On Fri, Jan 15, 2021 at 03:28:16AM -0800, Philipp Kern wrote:
> While you may have been careful now, the next one needs to stick to
> the same contract and the shell as provided by alpine (and by
> extension dash, the other default non-bash shell) is going to be the
> reference. Or is there a static checker that could provide us some
> assertion that a script might be POSIXly compliant?

I suppose there are a few options to tests shell scripts:

 -n (noexec): If not interactive, read commands but do not execute them.
   This is useful for checking the syntax of shell scripts.
 - checkbashisms: check whether a /bin/sh script contains any common
   bash-specific constructs (would complain about RANDOM).
 - shellcheck: lint tool for shell scripts

-- 
Valentin

pkern:
My fear was that tools like shellcheck might be specific to bash, but it turns out they have a /bin/sh shebang-specific mode as well. So I'm sort of fine with this if we have something like shellcheck (or checkbashisms, I didn't look into that) running as a presubmit.

burgerdev:
Thanks for the suggestions, I added a `shellcheck` step to ensure POSIX conformance during CI.

cli/test.sh
```diff
@@ -15,39 +15,43 @@
...
+trap cleanup EXIT
 
 for side in 0 1; do
         "$binary" genkey | tee "${t}/${side}" | "$binary" pubkey >"${t}/${side}.pub"
 done
 
+errors=0
 for counter in $(seq 0 "$iterations"); do
-  msg="$RANDOM"
+  msg="$(head -c 2 /dev/urandom | od -t u2 -A n)"
```
pkern, line 43: I think the od is not obvious and could use a comment.

pkern (commented):
LGTM with one nit.

cli/test.sh
burgerdev: I replaced that message generator with a static string because of (a) determinism and (b) bugfix (described in the commit message).

pkern (approved):
LGTM

PR #80, opened 2021-01-22
Apply clang-format to all C source files.

burgerdev:
The amount of source files keeps growing and it's easy to forget to add
them to the integration test script (case in point, happened to the pam
files). Since we do not include any third-party sources, the easiest
thing is to check all .c and .h files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,21 +58,10 @@ jobs:
         export DEBIAN_FRONTEND=noninteractive
         apt-get update
         apt-get install -y clang-format
-    - name: check libglome format
+    - name: check format
       run: |
-        clang-format --Werror --dry-run --style=google \
-          glome.c glome.h glome_test.c
-    - name: check glome-cli format
-      run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -58,21 +58,10 @@ jobs:
...
-        clang-format --Werror --dry-run --style=google \
-          cli/*.c cli/*.h
-    - name: check glome-login format
-      run: |
-        clang-format --Werror --dry-run --style=google \
-          login/config.c login/crypto.c login/crypto_test.c login/lockdown.c \
-          login/login.c login/login_test.c login/main.c login/ui.c \
-          login/base64.h login/config.h login/crypto.h login/lockdown.h \
-          login/login.h login/ui.h
+        find login -name "*.c" -or -name "*.h" | \
```
vvidic, line 63: It seems this would only check login directory now?

burgerdev: Thanks, fixed it.

vvidic (approved):

PR #82, opened 2021-01-26
Prevent random tag collisions in cli test.

burgerdev:


pkern (approved):
LGTM, although we should fix the randomness ONCE AND FOR ALL at some future point. ;)

PR #83, opened 2021-02-18
Test CLI with vectors from specification.

### LINKED ISSUE #50: CLI should be tested according to specification
We only test that a tag produced by the CLI is in turn correctly verified by the CLI - this fails to detect symmetric breakage, e.g. due to invalid parsing / reading. There should be a suite that verifies the test vectors from the specification.

burgerdev:
Fixes #50.

### DIFF
--- a/cli/test.sh
+++ b/cli/test.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-# Copyright 2020 Google LLC
+# Copyright 2021 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


vvidic (approved):
Looks good, perhaps only use hex values (`\xHH`) in `printf` instead of octal?

burgerdev:
Thanks for the review! Afaiu octal is the only portable binary encoding for `printf`: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/printf.html#tag_20_94_13.

vvidic:
Yes, I see now that BSD for example does not have hex in printf(1).

PR #86, opened 2021-04-20
Replace libconfuse with glib Key-value file parser.

### LINKED ISSUE #28: Rewrite glome-login's config module to use glib
Now that we agreed to use glib, we might want to consider moving away from libconfuse to something more solid/understandable like [glib's Key-value file parser](https://developer.gnome.org/glib/stable/glib-Key-value-file-parser.html).

vvidic:
Fixes #28.


pkern (approved):
I like. LGTM. Thank you!

PR #88, opened 2021-04-24
Run CI on ubuntu-latest

l9i:
ubuntu-latest points at 20.04 as of April 24, 2021.


vvidic (approved):
LGTM. Ubuntu 18 was getting a bit dated by now.

Issue #91, opened 2021-05-28
MVP: Login PAM module plus signing CLI


pkern:
We should actually go and design how glome-login is supposed to be used - with integration into Debian/Ubuntu as an example. I'd suggest the following:

* Users should be able to install a libpam-glome package and the distro scripts would add `pam_glome` to at least the `login` PAM `auth` chain, if not `common-auth`.
* This would do nothing (no failure or error message, `PAM_AUTHINFO_UNAVAIL`?) unless `/etc/glome/config` is found and contains both a service key and a URL prefix - or alternative both are passed via the arguments to the PAM module.
* The PAM module should use `pam_get_authtok` to get the tag, which honors `try_first_pass`. That way we can stack modules properly.
* `pam_glome` should then sort before `pam_unix` (in Debian/Ubuntu-speak its priority for `pam-auth-update` needs to be higher than `unix`'s 256 - `pam_unix` would then automatically gain `try_first_pass`). This would print the GLOME URL and a `Password:` prompt that would accept the tag. If the tag does not validate, PAM would continue with `pam_unix` and check the token as a password.
* This would of course namespace collide tags and passwords, but I think that's an acceptable tradeoff here. Users who do not want password validation can deactivate the password in `/etc/shadow` and `pam_unix` will DTRT.
* My understanding is that such a setup would work with regular terminal-like setups and even SSH - although very likely not with screensavers as the URL would not be displayed properly in that case. Note that the default timeout for `login` on Debian/Ubuntu is 60s - which is a little short for our protocol. We might want to document raising this through `/etc/login.defs`.
* We should have a minimal CLI that actually works with the glome-login protocol rather than just GLOME. The current CLI is kinda weird - it does not even generate the hex keys necessary for glome-login, so you first need to convert them away from binary.
* We could go with `glome login sign`/`tag` - or `glome-login sign`, I don't really care. But it also turns out that we do not have URL decomposition in C yet, which would be somewhat useful to feed URLs into. Basically I think it'd be great to have a CLI that takes the private key file as an argument and parses everything from the `/v1/` to the end and generates the appropriate tag to login with. Additionally there should be commands alike `wg genkey` and `wg pubkey` that DTRT. Then we'd be at the same technical level as Wireguard, where you still need to put those keys into the appropriate places in the config file - although we'd still ship an example config with the keys commented out. Technically this could also be written in a language different from C. I think from a distro point of view I'd weakly prefer it to be Python or C rather than Go.
[...]

PR #92, opened 2021-05-29
PAM: Check the authcode against the authtok

pkern:
Print the link to GLOME's login server through the conversational
interface (PAM_TEXT_INFO) and then check the authcode against the
(possibly cached) authtok. This allows proper chaining of PAM modules,
including password fallback through pam_unix. (#91)

Behavior change: This no longer prints the authcode while typing,
as it reuses the regular (echo off) password prompt.

### DIFF
--- a/login/pam.c
+++ b/login/pam.c
@@ -149,15 +149,41 @@ static int glome_authenticate(pam_handle_t *pamh, login_config_t *config,
   free(action);
   action = NULL;
 
-  char *input = NULL;
-  if (pam_prompt(pamh, PAM_PROMPT_ECHO_ON, &input,
-                 "%s%s\nEnter GLOME code obtained from the above link: ",
-                 config->url_prefix, url)) {
-    return failure(EXITCODE_PANIC, error_tag, "pam-prompt");
+  struct pam_conv *conv;
+  if (pam_get_item(pamh, PAM_CONV, (const void **)&conv) != PAM_SUCCESS) {
+    return failure(EXITCODE_PANIC, error_tag, "pam-get-conv");
   }
 
+  const char *template = "GLOME link: %s%s";
+  size_t message_len =
+      strlen(template) + strlen(config->url_prefix) + strlen(url) - 4 + 1;
+  char *message = malloc(message_len);
+  if (message == NULL) {
+    return failure(EXITCODE_PANIC, error_tag, "malloc-message");
+  }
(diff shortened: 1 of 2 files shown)


login/pam.c
```diff
@@ -149,16 +149,37 @@ static int glome_authenticate(pam_handle_t *pamh, login_config_t *config,
...
-    return failure(EXITCODE_PANIC, error_tag, "pam-prompt");
+  struct pam_conv *conv;
+  if (pam_get_item(pamh, PAM_CONV, (const void **)&conv) != PAM_SUCCESS) {
+    return failure(EXITCODE_PANIC, error_tag, "pam-get-conv");
   }
 
+  const char *template = "GLOME link: %s%s";
+  size_t message_len =
+      strlen(template) + strlen(config->url_prefix) + strlen(url) + 1;
+  char *message = malloc(message_len);
```
vvidic, line 160: Should we check and free the `malloc` return value?

login/pam.c
```diff
@@ -149,16 +149,37 @@ static int glome_authenticate(pam_handle_t *pamh, login_config_t *config,
...
-                 config->url_prefix, url)) {
-    return failure(EXITCODE_PANIC, error_tag, "pam-prompt");
+  struct pam_conv *conv;
+  if (pam_get_item(pamh, PAM_CONV, (const void **)&conv) != PAM_SUCCESS) {
+    return failure(EXITCODE_PANIC, error_tag, "pam-get-conv");
   }
 
+  const char *template = "GLOME link: %s%s";
+  size_t message_len =
+      strlen(template) + strlen(config->url_prefix) + strlen(url) + 1;
```
vvidic, line 159: Length includes `%s%s` here, so maybe subtract that?

login/pam.c
pkern: Yeah, we should. Especially the free-ing. :facepalm: Oh C, how I love thee. Done.

login/pam.c
pkern: I sort of hoped I'd get buy with overestimating. But ok, done.

vvidic (approved):

login/pam.c
l9i: Could you add a comment explaining where `4` came from (and perhaps another one next to `template` so that it taken into account when the template changes)?

PR #94, opened 2021-05-29
login: Fix compiler warnings

pkern:
- Fix an off-by-one error when writing the terminating null byte
  to the end of the buffer in get_hostname.
- GCC wants strncpy to refer to the length of the destination buffer
  and cannot track separate strlen-based memory allocations
  (https://gcc.gnu.org/bugzilla//show_bug.cgi?id=88059). As what we are
  implementing is effectively a strdup, let's replace it with the
  correct function.


l9i (approved):
Thank you!

PR #95, opened 2021-05-29
Refactor CI to allow for more modularized presubmit checks

l9i:


pkern:
I think it'd be useful to populate the workflow name field, as otherwise it's shown as the filename in the check overview.

I also don't understand the Python matrix. We should not test Python 2 (discontinued). And the matrix check then requests 3.x, so all tests on all platforms are executed with 3.9 instead of the matrix version. A sane pick of versions should also reduce the amount of checks.

l9i:
> I think it'd be useful to populate the workflow name field, as otherwise it's shown as the filename in the check overview.

Done. What convention would you prefer `${lang}-presubmit`, or `presubmit-${lang}`?

> I also don't understand the Python matrix. We should not test Python 2 (discontinued). And the matrix check then requests 3.x, so all tests on all platforms are executed with 3.9 instead of the matrix version. A sane pick of versions should also reduce the amount of checks.

Done. Any hints on what would be the sane choice of versions would be appreciated.

pkern:
Not speaking against this change, but we will also need to address internal bug b/182772536 while we are at it as the Copybara GitHub check references presubmits by their name (and also wants an exhaustive list for some reason).

I guess alternatively we can discuss again making GitHub primary, which might resolve that particular concern. /cc @burgerdev

pkern (approved):
Maybe we can just try it and fix up our internal presubmits later. :>

LGTM.

pkern:
Hm, the internal review tool uncovered some stray end of the line whitespace in some of the yaml files. Would you mind purging that?

l9i:
> Hm, the internal review tool uncovered some stray end of the line whitespace in some of the yaml files. Would you mind purging that?

Sure. I've also added the yamllint workflow.

PR #99, opened 2021-07-02
Remove unneccessary username limit.

burgerdev:
The username is passed to shell_action(), which allocates a buffer
dynamically and does not depend on the username being a particular size.

As a side-effect, this simplifies config processing by removing the
config validation step.


burgerdev:
@l9i

l9i (approved):
Very nice!

PR #102, opened 2021-07-27
RFD 001: GLOME Login v2

burgerdev:
This is the (slightly changed) public version of my internal proposal to simplify the protocol and remove ambiguity.


docs/rfd/001.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+where HEXDIG should refer to a digit or an uppercase letter A-F. This matches the definition in https://url.spec.whatwg.org/#url-path-segment-string, which supposedly supersedes the RFC.
+
+Thus, define `EscapePathSegment` as a function that escapes all characters that are not unreserved, sub delimiters, `:` or `@`. See the Appendix for how this function can be implemented in some of the major programming languages.
+
+
+### Message Encoding
+
+Constructing a message takes three parameters: hostid, action and (optionally) hostid-type. The content to be embedded in the URL consists of 2 path segments: host-segment and action-segment.
```
l9i, line 77: What would be the benefits of restricting the action only to one URL path segment? To detect that the URL has not been truncated (provided that there is a trailing slash at the end)?

docs/rfd/001.md
```diff
@@ -0,0 +1,181 @@
...
+*   Use only action 'shell'.
+
+With a short hostname `glome` and omitting the domain, the URL looks like http://glome/v2/lNXP/2TZm3tVBle2tHCbb6GLfMm5cpy1UcaMv7218zHKdCMw=/host1.int.example.com/shell/ and is 94 characters long. The corresponding GLOME Login v1 variant has 94 characters as well.
+
+
+# Alternatives considered
+
+## URL Format
+
+*   Encode both public keys in one base64 segment.
```
l9i, line 108: I'd be in favour of keeping GLOME Login v1 behaviour of base64-encoding keys together. Separating them in a human-readable way (i.e., with `/`) brings no benefit to the user while increasing the URL size by two "fixed" characters (i.e., `/` and `=`). We might as well use 4 octets for the key prefix and further reduce the likelihood of a collision.

l9i (commented):
Given the sufficient prefix lenght (e.g., 24 bits), I am not sure if we care too much about collisions but they are potentially problematic since they cause persistent failures. I think it'd be an interesting property to at least make such failures non-deterministic. One possible solution to this problem is to use the ephemeral key (since it's different for each attempt) as an indicator of which bits of the server public key to compare (e.g., as an offset). While it still does not eliminate the problem of collisions, a simple retry should solve the issue. That said, I am not convinced if the added complexity is worth it though.

The other, unrelated question is whether we want to detect if an URL has been garbled between the client and the server to let the user know. We could add a checksum that does not even have to increase the URL size as it could be XOR-ed with the key prefix to save space. Again, that complicates the design and perhaps we don't need that.

docs/rfd/001.md
burgerdev: @l9i please have a look at my argument above.

burgerdev:
@l9i I added the changes that we discussed in today's meeting, ptal:

- `prefix-type` now disambiguates `prefix7` usage similar, but incompatible, to v1
- reflected that we don't want to build full URLs anymore
- added X25519 background to the LSB/MSB decision

PR #108, opened 2021-10-01
pyglome: more consise exception messages

l9i:
1. Make pylint happy by using f-strings.
2. Rephrase messages to be more consise.
3. Fix the comparison against MIN_TAG_LEN (ge, not gt).

### DIFF
--- a/python/pyglome/glome.py
+++ b/python/pyglome/glome.py
@@ -83,7 +83,7 @@ def _public_key_encode(public_key: x25519.X25519PublicKey):
 
 def _tag(msg: bytes, counter: int, key: bytes) -> bytes:
     if not 0 <= counter <= 255:
-        raise ValueError(f'tag counter (={counter}) must be within [0; 255]')
+        raise ValueError(f'tag counter (={counter}) must be within [0, 255]')
 
     message = bytes([counter]) + msg  # msg: N_x|M_n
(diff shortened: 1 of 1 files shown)


pkern (approved):

python/pyglome/glome.py
```diff
@@ -127,9 +126,10 @@ def __init__(self,
         else:
             my_public_key = my_private_key.public_key()
 
-        if not Glome.MIN_TAG_LEN < min_peer_tag_len <= Glome.MAX_TAG_LEN:
-            raise ValueError('min_peer_tag_len must be in range {}-{}'.format(
-                Glome.MIN_TAG_LEN, Glome.MAX_TAG_LEN))
+        if not Glome.MIN_TAG_LEN <= min_peer_tag_len <= Glome.MAX_TAG_LEN:
+            raise ValueError(
+                f'min_peer_tag_len (={min_peer_tag_len}) is not within '
+                f'({Glome.MIN_TAG_LEN}; {Glome.MAX_TAG_LEN}]')
```
burgerdev, line 132: Should the left brace be a bracket?

l9i: Yes, good catch. I think I've updated the message first, and only then spotted the comparison bug. And obviously forgot to update the message again.

PR #111, opened 2021-12-13
Use the same auth function for PAM and login binary

vvidic:


login/config.h
```diff
@@ -51,6 +51,9 @@ typedef struct glome_login_config {
 
   // Explicitly set host-id to use in the login request.
   const char* host_id;
+
+  // Private data for use by login operations (e.g. storing pam handle).
+  void* priv;
```
burgerdev, line 56: I'm not a big fan of this, it looks like an inversion of responsibilities. At the very least, I'd want to make this dedicated to a specific login method (i.e. `pam_t* pam_handle;  // context info iff using PAM`).

burgerdev (commented):
My gut feeling about this is that only the prompting should be configurable, whereas the printing and logging should be controlled by config flags. I also have not made up my mind yet whether this change makes it easier to maintain the code, which is the reason for the change - correct?

login/login.c
```diff
@@ -249,24 +255,90 @@ int request_url(const uint8_t service_key[GLOME_MAX_PUBLIC_KEY_LENGTH],
...
-  if (config->options & VERBOSE) {
-    errorf(
-        "debug: options: 0x%x\n"
-        "debug: username: %s\n"
-        "debug: login: %s\n"
-        "debug: auth delay: %d seconds\n",
-        config->options, config->username, config->login_path,
-        config->auth_delay_sec);
-  }
+static void login_printf(glome_login_config_t* config, const char* format,
```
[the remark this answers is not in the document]
vvidic: Yes, `ops->printf` was only used by `glome-login` to report errors so I replaced it with the `error` function now.

login/login.c
```diff
@@ -314,24 +391,31 @@ int login_run(glome_login_config_t* config, const char** error_tag) {
   free(action);
   action = NULL;
 
-  printf("Obtain the one-time authorization code from:\n%s%s\n",
-         config->url_prefix, url);
-
+  const char* template = "GLOME link: %s%s";
```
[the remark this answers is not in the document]
vvidic: Ok, the prompt is now set separately for each instance.

login/login.c
```diff
@@ -249,24 +255,90 @@ int request_url(const uint8_t service_key[GLOME_MAX_PUBLIC_KEY_LENGTH],
...
+  va_list args;
+  va_start(args, format);
+  vfprintf(stderr, format, args);
+  va_end(args);
+  fflush(NULL);
+}
+
+static void login_syslog(glome_login_config_t* config, int priority,
+                         const char* format, ...) {
   if (config->options & SYSLOG) {
```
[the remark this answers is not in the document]
vvidic: `ops` struct is `static const` so it cannot be changed at runtime, I think this is safer. Previously the check for syslog was sprinkled in login code and I thought it would be better to just do it in one place.

login/pam.c
```diff
@@ -95,132 +95,60 @@ static int get_username(pam_handle_t *pamh, glome_login_config_t *config,
...
-    return failure(EXITCODE_PANIC, error_tag, "broken-template");
-  }
-  free(url);
-  url = NULL;
+static void auth_syslog(glome_login_config_t *config, int priority,
+                        const char *format, ...) {
+  pam_handle_t *pamh = (pam_handle_t *)config->priv;
+  va_list argptr;
+  va_start(argptr, format);
+  pam_vsyslog(pamh, priority, format, argptr);
```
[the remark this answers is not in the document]
vvidic: I think it is required for logging from PAM modules and it should be the same as `pam_syslog` we used previously. The man page says:

```
STANDARDS
The pam_syslog and pam_vsyslog functions are Linux-PAM extensions.
```

login/config.h
vvidic: Yes, this is not great, but using `pam_t` in `config.h` would require PAM library for building `glome-login` binary too?

vvidic:
Yes, in general the idea here is to remove the code duplication that was introduced by the PAM module. The change also tries to preserve current behavior as much as possible.

login/pam.c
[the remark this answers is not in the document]
burgerdev: Yep, this was more of a drive-by comment unrelated to the PR. I'll check whether this compiles on BSD and open a bug if it does not.

login/login.c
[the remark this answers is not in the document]
burgerdev: Agreed.

vvidic:
Did not find a better way for this, so left the `ops` functions as before. The rest is cleaned up a bit so please check again.

login/config.h
vvidic: Removed `priv` pointer and pass `pamh` directly now.

vvidic:
Ok, the `ops` functions are now gone to, so I hope this looks better now :)

login/login.c
```diff
@@ -249,24 +219,109 @@ int request_url(const uint8_t service_key[GLOME_MAX_PUBLIC_KEY_LENGTH],
...
+}
+#endif
+
+int login_authenticate(glome_login_config_t* config, pam_handle_t* pamh,
+                       const char* prompt_format, const char** error_tag) {
   if (is_zeroed(config->service_key, sizeof config->service_key)) {
     return failure(EXITCODE_PANIC, error_tag, "no-service-key");
   }
+  if (config->url_prefix == NULL) {
+    return failure(EXITCODE_PANIC, error_tag, "no-url-prefix");
```
burgerdev, line 323: This fixes a bug: before the change, the URL would have been just `(null)/v1/....`. I'd personally have opted for treating an absent prefix as empty, but this works for me, too. Just wanted to make sure that this API change is deliberate.

login/login.c
```diff
@@ -291,6 +346,11 @@ int login_run(glome_login_config_t* config, const char** error_tag) {
     return EXITCODE_PANIC;
   }
 
+  if (config->options & VERBOSE) {
+    login_syslog(config, pamh, LOG_DEBUG, "host ID: %s", host_id);
+    login_syslog(config, pamh, LOG_DEBUG, "action: %s", action);
```
burgerdev, line 351: Why not write both in the same syslog call?

login/login.c
```diff
@@ -314,24 +374,30 @@ int login_run(glome_login_config_t* config, const char** error_tag) {
   free(action);
   action = NULL;
 
-  printf("Obtain the one-time authorization code from:\n%s%s\n",
-         config->url_prefix, url);
-
+  size_t message_len =
+      strlen(prompt_format) + strlen(config->url_prefix) + strlen(url) - 4 + 1;
```
burgerdev, line 378: I think it would be good to document somewhere why this works (the requirement for the format string is to contain precisely two `%s`?)

burgerdev (approved):
I only have minor things, looks good otherwise. Thanks for refactoring!

login/login.c
vvidic: Done.

login/login.c
vvidic: Added empty prefix.

login/login.c
vvidic: Added a comment to the description of the `login_authenticate` function.

login/login.c
vvidic: Yes, sorry, I missed that part somehow.

burgerdev (approved):

PR #115, opened 2022-02-03
login: don't prefix with "(null)" if no URL is configured

delroth:
External users might prefer authenticating using "glome cli" instead of
using a web service. This improves the UI a bit in this situation.


vvidic:
Yes, PR #111 that updates the auth code also fixes this problem.

burgerdev:
Fixed in #111.

PR #118, opened 2022-02-07
Fix handling of long authorization codes

### LINKED ISSUE #116: glome-login leaving an extra newline in the input buffer before returning?
A tiny nit I've had with glome-login for a while (this behavior has been here for 1+ year): after I enter my authentication code and I get logged in, it seems like there is a spurious newline still in the input buffer which causes my shell to give me two prompts:

```
smol login: root
Obtain the one-time authorization code from:
<snip>
> <snip>
Authorization code: OK

[root@smol:~]# 

[root@smol:~]# 
```

vvidic:
glome-login will only read code up to a predefined size and leave the rest of the line in the input buffer for the shell to read.

Fixes #116.

### DIFF
--- a/login/login.c
+++ b/login/login.c
@@ -243,33 +243,44 @@ void login_syslog(glome_login_config_t* config, pam_handle_t* pamh,
   }
 }
 
-// read_stdin reads printable characters from stdin into buf. It returns:
+// read_stdin reads characters from stdin into buf. It returns:
 // -1, if it encounters an error while reading
-// -2, if it encounters EOF
+// -2, if it encounters invalid characters in the input
 // (buflen-1) if it read buflen-1 characters
 // <(buflen-1), if a newline was read before the buffer was full
 // If the return value is >=0, the buf is NULL-terminated.
+// Additionally, stdin is always advanced up to a newline (or EOF)
+// to prevent excess input from being read by a future shell process.
 static int read_stdin(char* buf, int buflen) {
-  int i = 0;
+  // Return error if we got no characters.
+  if (fgets(buf, buflen, stdin) == NULL) {
+    perror("ERROR when reading from stdin");
+    return -1;
+  }
 
-  while (i < buflen - 1) {
-    int n = read(STDIN_FILENO, buf + i, 1);
-    if (n < 0) {  // error while reading from stdin
-      perror("ERROR when reading from stdin");
-      return -1;
-    }
-    if (n == 0) {  // EOF
+  bool newline = false;
+  int len = strlen(buf);
+  if (buf[len - 1] == '\n') {
+    newline = true;
+    buf[len - 1] = '\0';
+    len--;
+  }
+
+  // Return error if we got a non-printable character.
+  for (int i = 0; i < len; i++) {
+    if (buf[i] < 0x20 || buf[i] > 0x7e) {
+      errorf("ERROR invalid characters read from stdin\n");
       return -2;
     }
-    if (buf[i] == '\n' || buf[i] == '\r') {  // newline
-      break;
-    } else if (buf[i] >= 0x20 && buf[i] <= 0x7e) {
-      // Advance the buffer pointer only if we got a printable character.
-      i++;
+  }
+
+  // Read stdin until a newline to avoid passing junk to shell.
+  if (!newline) {
+    for (int c = 0; c != EOF && c != '\n'; c = fgetc(stdin)) {
     }
   }
-  buf[i] = '\0';
-  return i;  // number of characters in the buffer without the NUL byte
+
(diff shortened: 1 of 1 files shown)


login/login.c
```diff
@@ -251,8 +251,9 @@ void login_syslog(glome_login_config_t* config, pam_handle_t* pamh,
 // If the return value is >=0, the buf is NULL-terminated.
 static int read_stdin(char* buf, int buflen) {
```
burgerdev, line 252: @l9i do you know why we have an `fgets(3p)` implementation in GLOME?

l9i: It's almost like `fgets` but not quite. The main difference is here: https://github.com/google/glome/blob/5121a2e4657119847729144a62385064c0f79773/login/login.c#L266-L269

`read_stdin` tries to ignore all spurious characters that may appear on a real serial console. One could argue that perhaps a similar effect could be achieved with `fgets` and sanitizing the outbut buffer afterwards (and repeating `fgets` if necessary). I'd be totally fine with changing the implementation or even making this behaviour optional.

burgerdev: I see. What kind of characters would we expect to spuriously appear on the console? I believe we need to rely on a `\n` indicating the end of the tag to fix #116.

vvidic: Based on the discussion on the meeting, updated to use `fgets` and return an error if non-printable characters are found. Also always read until the newline to avoid passing input to shell.

login/login.c
```diff
@@ -243,33 +243,45 @@ void login_syslog(glome_login_config_t* config, pam_handle_t* pamh,
...
+// read_stdin reads characters from stdin into buf. It returns:
 // -1, if it encounters an error while reading
-// -2, if it encounters EOF
 // (buflen-1) if it read buflen-1 characters
 // <(buflen-1), if a newline was read before the buffer was full
 // If the return value is >=0, the buf is NULL-terminated.
 static int read_stdin(char* buf, int buflen) {
-  int i = 0;
+  // Return error if we got no characters.
+  if (fgets(buf, buflen, stdin) == NULL) {
```
burgerdev, line 253: We will need to adjust the calling code to allocate more than `ENCODED_BUFSIZE(GLOME_MAX_TAG_LENGTH)` to account for the 0 byte, but that bug has been here before.

login/login.c
```diff
@@ -243,33 +243,45 @@ void login_syslog(glome_login_config_t* config, pam_handle_t* pamh,
...
+  }
 
-  while (i < buflen - 1) {
-    int n = read(STDIN_FILENO, buf + i, 1);
-    if (n < 0) {  // error while reading from stdin
-      perror("ERROR when reading from stdin");
+  // Return error if we got a non-printable character.
+  for (int i = 0; i < len; i++) {
+    if (buf[i] < 0x20 || buf[i] > 0x7e) {
+      perror("ERROR invalid characters read from stdin");
```
burgerdev, line 269: This should not use `perror`, but `errorf`.

login/login.c
```diff
@@ -243,33 +243,45 @@ void login_syslog(glome_login_config_t* config, pam_handle_t* pamh,
...
-    } else if (buf[i] >= 0x20 && buf[i] <= 0x7e) {
-      // Advance the buffer pointer only if we got a printable character.
-      i++;
+  }
+
+  // Read stdin until a newline to avoid passing junk to shell.
+  if (!newline) {
+    int c;
+    while ((c = fgetc(stdin)) != EOF) {
+      if (c == '\n') {
```
burgerdev, line 278: Could that be in the loop condition?

login/login.c
```diff
@@ -243,33 +243,45 @@ void login_syslog(glome_login_config_t* config, pam_handle_t* pamh,
   }
 }
 
-// read_stdin reads printable characters from stdin into buf. It returns:
+// read_stdin reads characters from stdin into buf. It returns:
```
burgerdev, line 246: Could you please update the docstring to say that we're always advancing stdin to a newline or EOF?

login/login.c
vvidic: It seems not, `ENCODED_BUFSIZE` already contains `+1` for the 0 byte, so that should work fine.

login/login.c
vvidic: Done.

login/login.c
vvidic: Done.

login/login.c
vvidic: Done.

login/login.c
burgerdev: I keep being surprised by that API.

burgerdev (approved):

PR #119, opened 2022-02-26
Make tests optional using a global meson option

vvidic:


login/meson.build
```diff
@@ -72,19 +73,21 @@ if get_option('pam-glome')
...
-            build_by_default : true, output : [ 'pam_service' ],
-            command : [ 'mkdir', '@OUTPUT@' ])
-        test('pam test', pam_test,
-            env: [ 'LD_PRELOAD=libpam_wrapper.so', 'PAM_WRAPPER=1',
-                'PAM_WRAPPER_SERVICE_DIR=' +
-                join_paths(meson.build_root(), 'login', 'pam_service'),
-                'PAM_GLOME=' +
-                join_paths(meson.build_root(), 'login', 'pam_glome.so') ])
+    if get_option('tests')
+        libpamtest = dependency('libpamtest', required : false)
```
pkern, line 77: Should we require libpamtest when PAM is wanted and tests are requested?

pkern (approved):
LGTM

login/meson.build
vvidic: Sure, let's give it a try :)

vvidic: Ah, no good, Alpine does not have this library. But at least Debian now runs this test ;)

pkern:
Thanks!

PR #120, opened 2022-02-27
Use OpenSSL INI parser to load configuration

vvidic:


vvidic:
The only downside to this is that `NCONF_get_section_names` we might need for MPA is only available in very recent OpenSSL 3.0.

vvidic:
PR updated to work with BoringSSL.

vvidic:
Closing as we are probably not going to use this.

PR #124, opened 2022-03-29
Run the C compilation presubmit on Debian stable and testing

pkern:
We saw breakage on internal presubmits against Debian testing.
debian:latest points to the latest Debian stable. Let's make that
part explicit.

This change depends on #123 being merged first to fix
compatibility with Debian testing.

### DIFF
--- a/.github/workflows/presubmit-c.yml
+++ b/.github/workflows/presubmit-c.yml
@@ -19,14 +19,15 @@ jobs:
     strategy:
       matrix:
         container:
-        - 'debian:latest'
+        - 'debian:stable'
+        - 'debian:testing'
         - 'alpine:latest'
     container:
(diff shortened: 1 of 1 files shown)


.github/workflows/presubmit-c.yml
```diff
@@ -19,7 +19,8 @@ jobs:
     strategy:
       matrix:
         container:
-        - 'debian:latest'
+        - 'debian:stable'
+        - 'debian:testing'
```
burgerdev, line 23: I think you need to change the if-clause on line 30 to use https://docs.github.com/en/actions/learn-github-actions/expressions#startswith.

pkern: Whoops, done.

burgerdev (approved):

PR #128, opened 2022-04-18
pam_glome: support fake passwords from OpenSSH

l9i:
OpenSSH provides PAM modules with a fake passwords when it is not going to permit the user to log in (e.g., due to PermitRootLogin=no), as per https://github.com/openssh/openssh-portable/commit/283b97ff33ea2c64116

Let `pam_glome` spare users debugging why valid authcodes are being rejected, by hinting at OpenSSH.

### DIFF
--- a/login/pam.c
+++ b/login/pam.c
@@ -168,7 +168,8 @@ int login_prompt(glome_login_config_t *config, pam_handle_t *pamh,
   // OpenSSH provides fake password when login is not allowed,
   // for example due to PermitRootLogin set to 'no'
   // https://github.com/openssh/openssh-portable/commit/283b97
-  const char fake_password[] = "\b\n\r\177INCORRECT"; // auth-pam.c from OpenSSH
+  const char fake_password[] =
+      "\b\n\r\177INCORRECT";  // auth-pam.c from OpenSSH
   bool is_fake = true;
 
   // Constant-time comparison in case token contains user's password
--- a/login/pam.c
+++ b/login/pam.c
@@ -164,6 +164,22 @@ int login_prompt(glome_login_config_t *config, pam_handle_t *pamh,
   if (strlen(token) >= input_size) {
     return failure(EXITCODE_PANIC, error_tag, "pam-authtok-size");
   }
+
+  // OpenSSH provides fake password when login is not allowed,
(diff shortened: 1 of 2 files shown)


login/pam.c
```diff
@@ -164,6 +164,23 @@ int login_prompt(glome_login_config_t *config, pam_handle_t *pamh,
...
+      "\b\n\r\177INCORRECT";  // auth-pam.c from OpenSSH
+  bool is_fake = true;
+
+  // Constant-time comparison in case token contains user's password
+  for (size_t i = 0; i < strlen(token); i++) {
+    is_fake &= (token[i] == fake_password[i % (sizeof(fake_password) - 1)]);
+  }
+  if (is_fake) {
+    pam_syslog(pamh, LOG_ERR, "login disallowed by OpenSSH (PermitRootLogin?)");
+    return EXITCODE_PANIC;
```
vvidic, line 181: To match the rest of the function, should this be:

return failure(EXITCODE_PANIC, error_tag, "pam-authtok-fake");

l9i: I had something like that initially but then I wanted to have something more user friendly.

PTAL

vvidic (approved):

login/pam.c
vvidic: Looks good.

login/pam.c
```diff
@@ -164,6 +164,22 @@ int login_prompt(glome_login_config_t *config, pam_handle_t *pamh,
...
   }
+
+  // OpenSSH provides fake password when login is not allowed,
+  // for example due to PermitRootLogin set to 'no'
+  // https://github.com/openssh/openssh-portable/commit/283b97
+  const char fake_password[] =
+      "\b\n\r\177INCORRECT";  // auth-pam.c from OpenSSH
+  bool is_fake = true;
+
+  // Constant-time comparison in case token contains user's password
```
pkern, line 175: I guess there's no advantage of moving the constant-time comparison function generically into crypto.c, right? (Alongside is_zeroed.)

l9i: Note that this comparison function is "special" as it loops around fake_password so that it matches the lenght of token's lenght. I don't have a strong preference either way.

PR #129, opened 2022-04-18
Add support for long and config options

vvidic:


login/config.c
```diff
@@ -101,48 +103,182 @@ static void key_value(char *line, char **key, char **val) {
...
+  } else {
+    return status_createf("ERROR: unrecognized boolean value: %s", val);
+  }
+}
+
+static status_t assign_key_option(uint8_t *dest, size_t dest_len,
+                                  const char *val) {
+  if (is_zeroed(dest, dest_len)) {
+    if (decode_hex(dest, dest_len, val)) {
+      return status_createf("ERROR: Failed to hex decode service key: %s", val);
```
pkern, line 194: We don't seem to be consistent in `ERROR: <uppercase>` vs. `ERROR: <lowercase>`.

login/pam.c
```diff
@@ -37,49 +37,43 @@ static int parse_pam_args(pam_handle_t *pamh, int argc, const char **argv,
                           glome_login_config_t *config) {
   memset(config, 0, sizeof(glome_login_config_t));
   int errors = 0;
+  status_t status;
 
   for (int i = 0; i < argc; ++i) {
     if (!strncmp(argv[i], "config_path=", 12)) {
```
pkern, line 43: I'm not saying that it'd be better readable but the offsets make me all queasy. Would it make sense to replace the string constants with `const char kConfigPath[] = "config_path=";` like constants and use sizeof on the constant instead of the hardcoded numbers?

login/ui.c
```diff
@@ -164,6 +195,11 @@ int parse_args(glome_login_config_t* config, int argc, char* argv[]) {
       default:
         return -1;  // PANIC
     }
+    if (status != STATUS_OK) {
+      errorf("%s", status);
```
pkern, line 199: I _think_ this misses a `\n`.

pkern (commented):
I like this. I mostly have some nits. Some of them are pre-existing, though.

login/config.c
vvidic: Done.

login/pam.c
vvidic: Sounds good, I added a helper function to handle this.

login/ui.c
vvidic: Thanks!

pkern (approved):
LGTM, thanks a lot!

PR #130, opened 2022-05-16
Sync PAM config options names with the rest of the code.

### LINKED ISSUE #127: PAM module arguments naming is inconsistent with the config
It would be great if PAM module arguments naming was consistent with the config (and flags in the future once glome-login suports long flags).

Examples of inconsistency:
- `service_key` (PAM) vs `key` (config)
- `service_key_version` (PAM) vs `key-version` (config)
- `url_prefix` (PAM) vs `url-prefix` (config)

Personally, I like the config naming better. It's shorter and uses dashes instead of underscores.

vvidic:
Use the same parameter names as in the config file and long command line
options but allow underscore instead of dash.

Fixes #127.

### DIFF
--- a/login/pam.c
+++ b/login/pam.c
@@ -33,10 +33,20 @@
 
 #define UNUSED(var) (void)(var)
 
-static const char *arg_value(const char *arg, const char *key) {
-  int key_len = strlen(key);
-  if (strncmp(arg, key, key_len) == 0 && arg[key_len] == '=') {
+static const char *arg_value(const char *arg, const char *key,
+                             const char *default_value) {
+  int i, key_len = strlen(key);
+  for (i = 0; i < key_len; i++) {
+    // Compare key with arg char by char while also allowing _ in place of -
+    if (!(key[i] == arg[i] || (key[i] == '-' && arg[i] == '_'))) {
+      return NULL;
+    }
+  }
+  if (arg[key_len] == '=') {
     return arg + key_len + 1;
+  }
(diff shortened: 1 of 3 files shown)


login/pam.c
```diff
@@ -33,11 +33,20 @@
 
 #define UNUSED(var) (void)(var)
 
-static const char *arg_value(const char *arg, const char *key) {
-  int key_len = strlen(key);
-  if (strncmp(arg, key, key_len) == 0 && arg[key_len] == '=') {
+static const char *arg_value(const char *arg, const char *key,
+                             const char *default_value) {
+  int i, key_len = strlen(key);
+  for (i = 0; i < key_len; i++) {
```
pkern, line 39: Given that this is only bounded by `strlen(key)`, couldn't this overrun the end of the `argv` buffer?

login/pam.c
```diff
@@ -33,11 +33,20 @@
...
+    if (!(key[i] == arg[i] || (key[i] == '-' && arg[i] == '_'))) {
```
pkern, line 40: I think this line deserves a comment.

login/pam.c
vvidic: `arg` is a null-terminated string and if it is shorter than the `key` it will cause the if statement inside the loop to fail: null char in `arg` will not be equal to the regular char in `key`.

login/pam.c
vvidic: Done.

pkern (approved):

PR #135, opened 2022-10-09
Update CLI to use base64 instead of hex tags.

### LINKED ISSUE #121: CLI should handle tags as base64 instead of hex
It's slightly confusing that the CLI prints and verifies tags in hex format, while the tag is exchanged base64-encoded in the login protocol. base64 has arguably better utility support and would allow using the cli as client without re-encoding, so I propose changing the output of `glome tag` to base64.

vvidic:
README file is also updated as it was describing an older version of the CLI.

Fixes #121.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -7,35 +7,23 @@ This is a CLI utility to facilitate GLOME operations from the command line.
 Generating two key pairs:
 
 ```shell
-$ glome Alice >Alice.pub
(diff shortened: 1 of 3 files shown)


vvidic:
It says: _You’re not [authorized](https://docs.github.com/articles/about-protected-branches/) to merge this pull request._

So someone else will need to press Merge here :)

PR #142, opened 2022-12-11
Use project version in library version

vvidic:
Library becomes libglome.so.0.1 instead of libglome.so.0.


pkern:
Agreed in the meeting that we must not change the ABI/API of libglome during the 0.x cycle if we go with this or we need to touch the versioning again. But I guess it's better than the alternatives for now, in absence of more native support in meson to produce sensible version numbers for libraries.

PR #150, opened 2023-03-11
Use GitHub's super-linter

l9i:


burgerdev:
Looks like this is the little sibling of #151

PR #151, opened 2023-03-11
Use GitHub's super-linter

l9i:


pkern:
I can see some value in the concept but it also means that we are deferring any decision on how to lint the code to the incarnation of the super linter we reference. Could you link to how this is configured when we need to modify something? There are things in there were my mind just explodes (like referencing packages by version in apt-get called from Dockerfile - whyyyyy?, that just sounds brittle). There are likely also some good linters in here, but clearly we'd then also need to fix all the linting issues first. :/

pkern:
I think we agreed in the meeting that this is not the way to go, at least for now.

Issue #122, opened 2022-03-23
Support more fleet-wide settings in the config file


burgerdev:
It would be convenient if we could set a couple of options that are usually constant in a given deployment environment with the config file. That would allow sharing a `glome-login` binary between environments, and customizing each environment with the config file only. Things that I can think of:

- `hostid-type` (`hostid` would still be obtained from `gethostname(2)`)
- `syslog` config
- timeouts and sleeps
- minimum tag length

vvidic:
Yes, I was planning to add long command line options and also make them available from the config file.

burgerdev:
I would really like to see `hostid-type` support.

vvidic:
Yes, most options should be available in the config now now. So, let me try to add that minimum tag length and host id type.

vvidic:
I think all the mentioned options are added now, so closing this issue!

PR #153, opened 2023-03-12
Add config option for minimum tag length #122

vvidic:
Required tag length can only be increased from the default value up to maximum tag length.


login/login.c
```diff
@@ -445,6 +437,19 @@ int login_authenticate(glome_login_config_t* config, pam_handle_t* pamh,
     return failure(EXITCODE_INVALID_INPUT_SIZE, error_tag, "authcode-length");
   }
 
+  size_t min_len = MIN_ENCODED_AUTHCODE_LEN;
+  if (config->min_authcode_len > min_len) {
```
pkern, line 441: I think we should error out if a min authcode length larger than the tag length is being requested. (Because you'd never be able to satisfy that.)

vvidic: Added a check for that case.

login/login.c
```diff
@@ -428,12 +428,25 @@ int login_authenticate(glome_login_config_t* config, pam_handle_t* pamh,
...
 
-  if (bytes_read < MIN_ENCODED_AUTHCODE_LEN) {
+  size_t min_len = MIN_ENCODED_AUTHCODE_LEN;
+  if (config->min_authcode_len > min_len) {
+    if (config->min_authcode_len > strlen(authcode_encoded)) {
+      login_syslog(config, pamh, LOG_INFO,
+                   "minimum authcode too long: %d bytes (%s)",
+                   config->min_authcode_len, config->username);
+      login_error(config, pamh,
+                  "Minimum input too long: expected at least %d characters.\n",
```
pkern, line 438: at most?

pkern (approved):
LGTM

PR #154, opened 2023-03-12
Update Docker scripts for new public key format

vvidic:
Also updates the example config to use public-key instead of key.


pkern (approved):
LGTM

PR #156, opened 2023-03-13
Insert a slash after url-prefix when writing it into prompt

pkern:
We previously had URLs like "url-prefix = http://otac" that got /v1/[...] appended. With the recent changes to prompt the slash before v1 was dropped, which makes these URL prefixes invalid. As we are dropping url-prefix before v1 anyway, let's fork the config and add some more compatibility code to ease the migration to prompt.

### DIFF
--- a/login/config.c
+++ b/login/config.c
@@ -280,7 +280,16 @@ static status_t assign_service_option(glome_login_config_t *config,
   } else if (strcmp(key, "url-prefix") == 0) {
     // `url-prefix` support is provided only for backwards-compatiblity
     // TODO: to be removed in the 1.0 release
-    return assign_string_option(&config->prompt, val);
+    size_t len = strlen(val);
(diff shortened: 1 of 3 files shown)


login/config.c
```diff
@@ -280,7 +280,13 @@ static status_t assign_service_option(glome_login_config_t *config,
   } else if (strcmp(key, "url-prefix") == 0) {
     // `url-prefix` support is provided only for backwards-compatiblity
     // TODO: to be removed in the 1.0 release
-    return assign_string_option(&config->prompt, val);
+    size_t len = strlen(val);
+    char *url_prefix = malloc(len + 2);
+    strncpy(url_prefix, val, len + 1);
```
burgerdev, line 285: This should technically check for `malloc` problems.

burgerdev (approved):

PR #157, opened 2023-03-14
Parse command line again after reading the config

vvidic:
This way command line options can override the settings in the configuration file. Same goes for the PAM options.


pkern (approved):
LGTM

PR #169, opened 2023-08-10
Implement GLOME login INI config parsing in Go.

### LINKED ISSUE #168: Add Go GLOME config file parsing
We have an internal user of GLOME that's written their own config parser. Let's provide an in-tree config parser in Go, so we can include it in test suites publically and keep it updated in lockstep with the C one.

lukegb:
This implements INI parsing internally, rather than pulling in a library because
we want to exactly match the behaviour of the C library, and our INI-parsing has
a few quirks (as do all INI parsers...) - for instance, we only support comments
if they're on their own line.

Additionally, the code that does the INI parsing is relatively small; the bulk of
the code is actually parsing things into application specific structs, which we
would need to do anyway. Possibly this is ripe for switching to TOML, but we'd still
have to e.g. parse string encoded keys to binary/validate that the options set
are compatible with each other.

Fixes #168.

### DIFF
--- a/go/config/config.go
+++ b/go/config/config.go
@@ -1,0 +1,347 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"bufio"
+	"encoding/base64"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+	"unicode"
+
+	"github.com/google/glome/go/glome"
+)
+
+// Config represents the supported GLOME login settings.
+type Config struct {
+	AuthDelay      int
+	InputTimeout   int
+	ConfigPath     string
+	EphemeralKey   glome.PrivateKey
+	MinAuthcodeLen int
+	HostID         string
+	HostIDType     string
+	LoginPath      string
+	DisableSyslog  bool
+	PrintSecrets   bool
+	Timeout        int
+	Verbose        bool
+
+	ServiceConfig ServiceConfig
+}
+
+// ServiceConfig contains GLOME settings from the [service] configuration section.
+type ServiceConfig struct {
+	PublicKey  glome.PublicKey
+	KeyVersion int
+	Prompt     string
+}
+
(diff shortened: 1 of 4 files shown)


go/config/config_x_test.go
pkern, line 1: I find the naming of the file a little weird. I think I'd slightly prefer if this was just merged into the other file. This would also make it more obvious what of the parse logic is actually checked for.

go/config/config.go
```diff
@@ -0,0 +1,316 @@
...
+	var err error
+	switch key {
+	case "auth-delay":
+		err = interpretPositiveInt(value, &cfg.AuthDelay)
+	case "input-timeout":
+		err = interpretPositiveInt(value, &cfg.InputTimeout)
+	case "config-path":
+		cfg.ConfigPath = value
+	case "ephemeral-key":
+		err = interpretPrivateKey(value, hex.DecodeString, &cfg.EphemeralKey)
```
pkern, line 159: Can we yell loudly when this is set?

go/config/config.go
```diff
@@ -0,0 +1,316 @@
...
+		return "", "", fmt.Errorf("empty key is invalid")
+	}
+	line = line[keyEnd:]
+
+	// Value is the line starting from the first non-space after =.
+	valueStart := strings.IndexFunc(line, func(r rune) bool {
+		return !unicode.IsSpace(r) && r != '='
+	})
+	separator := line[:valueStart]
+	value = line[valueStart:]
```
pkern, line 231: This panics on "a =", I think?

go/config/config_x_test.go
```diff
@@ -0,0 +1,224 @@
...
+	}, {
+		name:      "invalid config line",
+		config:    "hello",
+		wantError: config.BadKeyValue,
+	}, {
+		name:      "invalid config line with spaces",
+		config:    "hello a b c",
+		wantError: config.BadKeyValue,
+	}, {
+		name:      "missing key",
```
pkern, line 120: + missing value ;-)

pkern (changes requested):
I overcame my struggles with having to maintain an INI parser implementation in the future. Turns out our config format feels very bespoke anyway.

go/config/config.go
lukegb: Now we require an AllowInsecureOptions parameter.

go/config/config.go
lukegb: Fixed (and tested)

go/config/config_x_test.go
lukegb: Done.

go/config/config_x_test.go
lukegb: Done.

burgerdev:
LGTM, but can we please keep track of the problems with our config format that prevent us from using an ini parser dependency? Next time we touch the config format, we could aim to fix those.

pkern (approved):
From my side this LGTM now. I'm with Markus that we shoud document somewhere (maybe in the commit description? or in a source comment...) why we chose this route. If it's just about the value parsing, we should've been able to do that also with another implementation. That said, Parse() is actually very small and the remainder is spent on putting the values into the right places, so this is fine with me for now.

PR #173, opened 2024-01-25
Avoid calling openlog when using GLOME as a library

lukegb:
This causes log lines to always use `glome-login` as the ident, which isn't necessarily desirable.

We can just not call openlog: if we don't call it, syslog will initialise with some default values (i.e. not including the PID), so this won't _break_ any users of GLOME-as-a-library who don't update to call `openlog`. To avoid this then logging to the wrong facility we update our own calls to syslog to explicitly set the facility.

PAM doesn't need a facility; it appears to just use LOG_AUTHPRIV internally.

### DIFF
--- a/login/main.c
+++ b/login/main.c
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <stdlib.h>
+#include <syslog.h>
 #include <unistd.h>
 
 #include "config.h"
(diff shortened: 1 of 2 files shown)


login/main.c
```diff
@@ -43,6 +44,11 @@ int main(int argc, char* argv[]) {
     return EXITCODE_PANIC;
   }
 
+  // Initialize syslog, if we're going to log.
+  if (config.options & SYSLOG) {
```
pkern, line 48: Do we support setting this in the config file? Then it'd need to be later. (Which is fine, it was done in run before anyway.)

lukegb: We do, we have `disable-syslog`. I'll update the PR.

pkern (approved):
LGTM

PR #176, opened 2024-03-07
Fix handling of prefix type in login CLI

vvidic:
Key index is the common case we want to support here:

If the MSB is set, the low bits are interpreted as a 7 bit integer, which the server should interpret as the index of the key its supposed to use.

### DIFF
--- a/cli/commands.c
+++ b/cli/commands.c
@@ -359,10 +359,18 @@ int login(int argc, char **argv) {
     fprintf(stderr, "handshake size is invalid in path %s\n", path);
     goto out;
   }
-  if ((handshake[0] & 0x80) != 0) {
-    fprintf(stderr,
-            "only \"service-key-indicator\" prefix type is supported\n");
-    goto out;
+  if ((handshake[0] & 0x80) == 0) {
+    uint8_t public_key[GLOME_MAX_PUBLIC_KEY_LENGTH] = {0};
+    if (glome_derive_key(private_key, public_key)) {
+      fprintf(stderr, "unable to generate a public key\n");
+      goto out;
+    }
+    // Most significant bit is not set for X25519 key (see RFC 7748).
+    uint8_t public_key_msb = public_key[GLOME_MAX_PUBLIC_KEY_LENGTH - 1];
+    if (handshake[0] != public_key_msb) {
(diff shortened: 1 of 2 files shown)


burgerdev (commented):
I believe the reasoning was that we only have one key at the command line, so we can't pick one by index and should rather just accept the prefix indicator. I'd also have expected an MSB check further down, but don't see it.

However, when I'm already despreate enough to use the CLI for verification, I probably don't care whether it's index or prefix, and rather want the tool to just produce a tag with the given key.

How about we accept any index without check, but do the MSB check when it's a prefix?

vvidic:
Updated as suggested, please check.

cli/commands.c
```diff
@@ -359,10 +359,17 @@ int login(int argc, char **argv) {
...
-            "only \"service-key-indicator\" prefix type is supported\n");
-    goto out;
+  if ((handshake[0] & 0x80) == 0) {
+    uint8_t public_key[GLOME_MAX_PUBLIC_KEY_LENGTH] = {0};
+    if (glome_derive_key(private_key, public_key)) {
+      fprintf(stderr, "unable to generate a public key\n");
+      goto out;
+    }
+    uint8_t public_key_msb = public_key[GLOME_MAX_PUBLIC_KEY_LENGTH - 1];
+    if ((handshake[0] & 0x7f) != (public_key_msb & 0x7f)) {
```
burgerdev, line 369: Nit: it should be sufficient to test `handshake[0] != public_key_msb`.

burgerdev (approved):
Thanks!

cli/commands.c
vvidic: Tnx, removed 0x7f from handshake, but not sure it is possible for the key? How do we know high bit is not set here?

burgerdev: X25519 public keys are specified to have their most significant bit cleared (https://www.rfc-editor.org/rfc/rfc7748#section-5 second paragraph). That peculiarity is what allowed @l9i to use that bit for metadata in the first place :)

vvidic: Ok, updated with a comment.

pkern (approved):
LGTM, thanks!

PR #177, opened 2024-03-13
Disable bracketed paste

pkern:
In some settings we seem to end up with users' terminals being confused and willing to send bracketed paste escape codes, which then cause authentication to fail. Even worse it looks like the special characters are not displayed in some shells, thus confusing users as to what just happened.

Although to be honest: I have not found the specific environment where this breaks in a reproducible fashion and this is effectively defence in depth.

### DIFF
--- a/login/login.c
+++ b/login/login.c
@@ -41,6 +41,7 @@
 #include "ui.h"
 
 #define PROMPT "> "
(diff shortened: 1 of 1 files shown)


lukegb (approved):
¯\_(ツ)_/¯

PR #182, opened 2024-06-20
Support for macOS

pkern:
This implements some source changes to make the code base compatible with macOS. Which is, of course, a little frustrating to fix. I have no idea how compatibility with other awkward OSes (like BSDs) will be after this change.

### DIFF
--- a/login/login.c
+++ b/login/login.c
@@ -12,10 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// For vsyslog
-#define _BSD_SOURCE
-#define _DEFAULT_SOURCE
-
 #include "login.h"
 
 #include <assert.h>
@@ -211,9 +207,15 @@ void login_syslog(glome_login_config_t* config, pam_handle_t* pamh,
                   int priority, const char* format, ...) {
   UNUSED(pamh);
   if (config->options & SYSLOG) {
+    char* buf = calloc(1024, 1);
+    if (!buf) {
+      return;
+    }
+
     va_list args;
     va_start(args, format);
-    vsyslog(LOG_MAKEPRI(LOG_AUTH, priority), format, args);
+    vsnprintf(buf, 1024, format, args);
+    syslog(priority, "%s", buf);
     va_end(args);
   }
 }
--- a/login/login.c
+++ b/login/login.c
@@ -308,7 +308,10 @@ int login_prompt(glome_login_config_t* config, pam_handle_t* pamh,
 static char* create_login_message(glome_login_config_t* config,
                                   pam_handle_t* pamh, const char** error_tag) {
   char* host_id = NULL;
-  int max_hostname_len = sysconf(_SC_HOST_NAME_MAX) + 1;
+  int max_hostname_len = sysconf(_SC_HOST_NAME_MAX);
+  if (max_hostname_len == -1) {
+    max_hostname_len = _POSIX_HOST_NAME_MAX;
+  }
 
   if (config->host_id != NULL) {
     host_id = strdup(config->host_id);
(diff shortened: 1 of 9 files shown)


login/pam.c
```diff
@@ -29,6 +33,11 @@
 
 #define MODULE_NAME "pam_glome"
 
+#ifndef PAM_GLOME_LINUX
+void pam_syslog(void *pamh, ...) { (void)(pamh); }
```
burgerdev, line 37: Should we implement this in terms of `syslog` instead?

burgerdev (approved):

burgerdev:
Maybe we should just abandon `vsyslog`.

```diff
diff --git a/login/login.c b/login/login.c
index 0f6689b..1f20804 100644
--- a/login/login.c
+++ b/login/login.c
@@ -12,10 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// For vsyslog
-#define _BSD_SOURCE
-#define _DEFAULT_SOURCE
-
 #include "login.h"
 
 #include <assert.h>
@@ -211,9 +207,15 @@ void login_syslog(glome_login_config_t* config, pam_handle_t* pamh,
                   int priority, const char* format, ...) {
   UNUSED(pamh);
   if (config->options & SYSLOG) {
+    char* buf = calloc(1024, 1);
+    if (!buf) {
+      return;
+    }
+
     va_list args;
     va_start(args, format);
-    vsyslog(LOG_MAKEPRI(LOG_AUTH, priority), format, args);
+    vsnprintf(buf, 1024, format, args);
+    syslog(priority, "%s", buf);
     va_end(args);
   }
 }
diff --git a/meson.build b/meson.build
index 64a51d4..a23b0b0 100644
--- a/meson.build
+++ b/meson.build
@@ -19,9 +19,7 @@ pkg = import('pkgconfig')
 
 sysconfdir = join_paths(get_option('prefix'), get_option('sysconfdir'))
 add_project_arguments('-DSYSCONFDIR="' + sysconfdir + '"', language : 'c')
-if host_machine.system() != 'darwin'
-    add_project_arguments('-D_POSIX_C_SOURCE=200809L', language : 'c')
-endif
+add_project_arguments('-D_POSIX_C_SOURCE=200809L', language : 'c')
 add_project_arguments('-DOPENSSL_API_COMPAT=10100', language : 'c')
 
 openssl_dep = dependency('openssl', version : '>=1.1')
```

login/login.c
```diff
@@ -308,6 +308,7 @@ int login_prompt(glome_login_config_t* config, pam_handle_t* pamh,
 static char* create_login_message(glome_login_config_t* config,
                                   pam_handle_t* pamh, const char** error_tag) {
   char* host_id = NULL;
+  int max_hostname_len = sysconf(_SC_HOST_NAME_MAX) + 1;
```
vvidic, line 311: Should we check return value of sysconf?

login/pam.c
vvidic: With `syslog` it might end up in a different log file or with different format than the other PAM modules so not sure about that.

vvidic (approved):

login/login.c
pkern: Fair enough, yes. Let's pick _POSIX_HOST_NAME_MAX (given that it's per standard the minimum, even if Linux disagrees) as the fallback value.

login/pam.c
pkern: To be fair: I have only made this compile and did not test against PAM's Darwin implementation. I suspect that way lies a lot of pain. My main worry here would be how to do it properly on the other BSD flavors.

For instance pam_krb5 documents these options:

* debug: turns on debugging via [syslog](https://linux.die.net/man/3/syslog)(3). Debugging messages are logged with priority LOG_DEBUG.
* debug_sensitive: turns on debugging of sensitive information via [syslog](https://linux.die.net/man/3/syslog)(3). Debug messages are logged with priority LOG_DEBUG.

Russ Albery wrote https://github.com/rra/pam-krb5/blob/main/portable/pam_vsyslog.c which I find amusing. Presumably we could handle compatibility the same way...

And there I was thinking that PAM is the proper abstraction between different OSes. ;-)

login/pam.c
```diff
@@ -29,6 +33,11 @@
...
+void pam_vsyslog(void *pamh, ...) { (void)(pamh); }
+#endif
```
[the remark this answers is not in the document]
pkern: I renamed the define to HAVE_PAM_EXT but made the include conditional on the else branch.

PR #203, opened 2025-02-26
go/login: provide a Stringer on the v1 GLOME URLResponse

lukegb:


pkern (changes requested):
`login/login.go:188:9: cannot use sb (variable of type strings.Builder) as string value in return statement` Hah, I wondered how that worked. Turns out it doesn't. :P

lukegb:
> `login/login.go:188:9: cannot use sb (variable of type strings.Builder) as string value in return statement` Hah, I wondered how that worked. Turns out it doesn't. :P

Should be fixed now... whoops.

pkern (approved):

PR #212, opened 2026-01-13
go/glome/glome_test: Use format strings for t.Fatalf()

pkern:
Otherwise we are using the error in a format string, which is a big no-no.


burgerdev (approved):
Thanks! Interesting - I'd have expected this to be flagged by the linter (if only for using `Fatalf` without arguments).

PR #221, opened 2026-03-26
Remove C CLI

### LINKED ISSUE #187: Remove C CLI
The Rust CLI should be the canonical implementation, because:

* It's quicker to iterate on features (e.g. adding a subcommand).
* It's safer (memory, types).

vvidic:
Keep README.md until we create a manpage for CLI.

Closes #187


kokoro/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/library/debian:bullseye AS build
+FROM docker.io/library/debian:latest AS build
```
burgerdev: Is using `latest` a good idea? I'd think we might want to be deliberate about updating the version here. For reproducibility, we might even want to pin the version/digest.

kokoro/rodete/debian_cargo.sh
```diff
@@ -0,0 +1,36 @@
...
+    ver=${ver%%-*}
+    sed -i "/^$1/s/= \"[.0-9]*\"/= \"$ver\"/" rust/Cargo.toml
+  fi
+}
+
+fix_version base64
+fix_version hex-literal
+fix_version tempfile
+fix_version yaml-rust2
+rm -f rust/Cargo.lock
```
burgerdev: Could we somehow align the versions in the `Cargo.lock` and the ones from Debian more permanently?

burgerdev (approved):
Nice, thanks!

kokoro/docker/Dockerfile
vvidic: Updated to trixie as the current stable version.

kokoro/rodete/debian_cargo.sh
vvidic: Updated versions in Cargo.toml so they work with current Debian packages.

vvidic:
Rebased the change so it should be possible to merge now. Please check again.

vvidic:
Reverted the change for docker build command. It still works but shows a warning that buildx is recommended now.

PR #222, opened 2026-03-31
CLI Truncated Tag Authentication fix

ikkebr:
The CLI `verify` command accepts truncated authentication tags of arbitrary length, including a single byte. The verification routine base64url-decodes the user-supplied `--tag` argument and uses the *decoded length* as the comparison length for `CRYPTO_memcmp()`. No minimum tag length is enforced. An attacker supplying a 1-byte tag only needs to match the first byte of the real tag.

### DIFF
--- a/cli/commands.c
+++ b/cli/commands.c
@@ -257,6 +257,10 @@ int verify(int argc, char **argv) {
   if (expected_tag_len > sizeof tag) {
     expected_tag_len = sizeof tag;
   }
+  if (expected_tag_len < sizeof tag) {
+    fputs("MAC tag verification failed\n", stderr);
+    goto out;
+  }
 
   // compare the tag
   if (CRYPTO_memcmp(expected_tag, tag, expected_tag_len) != 0) {


burgerdev:
Hi @ikkebr,

there may be a misunderstanding here, this is actually working as intended.

When using the CLI, you are in full control of the arguments you pass to it, and can verify any sort of requirements externally, keeping the (C) implementation simple. Anyway, forcing the full tag length is definitely not how it's intended to be used - you'd usually want to check something like 4 bytes. This is what's implemented in the `glome-login` CLI and in the PAM module, where we're directly dealing with attacker-supplied data and have no means to implement external checks.

You might be interested in the Rust CLI, which is about to replace the C CLI (https://github.com/google/glome/pull/221) and has an additional flag with which you can enforce a minimum length. 

https://github.com/google/glome/blob/d680822f0a70e844215790f7a931fb1f0274edfd/rust/src/cli/bin.rs#L42-L48

Cheers, Markus

PR #226, opened 2026-06-01
glome: improve public API documentation in glome.h

### LINKED ISSUE #167: Improve documentation of `glome.h`
Instead of single sentences, the public functions should document all their parameters, especially wrt ownership, and describe the scenario where the function is appropriate.

FreezB11:
## Summary

Improves the documentation of the three public functions in `glome.h` as requested in #167.

The existing comments were single-sentence descriptions that gave no guidance on parameter ownership, buffer requirements, or when to use each function. This change replaces them with detailed doc comments covering all of that.

### `glome_generate_key`
- Describes the appropriate use case: ephemeral keys per session vs. long-lived service keys
- Documents both output buffers, their size requirements, and that the caller owns them and is responsible for zeroing the private key after use

### `glome_derive_key`
- Clarifies when to use it: recovering the public half of a previously stored private key
- Makes explicit that the function does not modify or retain the private key it receives

### `glome_tag`
- Explains the underlying mechanism: HMAC-SHA256 over a Diffie-Hellman derived shared secret
- Describes both operating modes separately: tag generation (`verify=false`) for the sender and tag verification (`verify=true`) for the receiver
- Documents every parameter individually, including the ownership direction change for `tag` depending on the value of `verify`
- Clarifies return value semantics for each mode

Closes #167.


glome.h
```diff
@@ -11,32 +11,89 @@
...
 #define GLOME_H_
-
 #include <stdbool.h>
 #include <stdint.h>
 #include <unistd.h>
-
 #define GLOME_MAX_PUBLIC_KEY_LENGTH 32
 #define GLOME_MAX_PRIVATE_KEY_LENGTH 32
 #define GLOME_MAX_TAG_LENGTH 32
-
```
burgerdev, line 25: Please consider reverting whitespace changes here.

glome.h
```diff
@@ -11,32 +11,89 @@
...
+//   private_key - Output buffer of at least GLOME_MAX_PRIVATE_KEY_LENGTH bytes
+//                 that receives the newly generated private key. The caller
+//                 owns this buffer and is responsible for zeroing and freeing
+//                 it when no longer needed.
+//   public_key  - Output buffer of at least GLOME_MAX_PUBLIC_KEY_LENGTH bytes
+//                 that receives the corresponding public key. The caller owns
+//                 this buffer.
+//
+// Returns 0 on success, or a negative value if key generation fails (e.g. the
+// underlying random number source is unavailable).
```
burgerdev: The `err` value is a bunch of comparisons OR'd together - I'm not too confident that it will always be negative on error. Let's just say non-zero instead.

glome.h
```diff
@@ -11,32 +11,89 @@
...
+//
+// Parameters:
+//   private_key - Input buffer of at least GLOME_MAX_PRIVATE_KEY_LENGTH bytes
+//                 containing the private key. The caller retains ownership;
+//                 this function does not modify or store the private key.
+//   public_key  - Output buffer of at least GLOME_MAX_PUBLIC_KEY_LENGTH bytes
+//                 that receives the derived public key. The caller owns this
+//                 buffer.
+//
+// Returns 0 on success, or a negative value if derivation fails.
```
burgerdev: Dito

glome.h
```diff
@@ -11,32 +11,89 @@
...
+//
+// Use verify=false (tag generation) on the sending side to produce a tag that
+// authenticates a message to the remote peer. Use verify=true (tag
+// verification) on the receiving side to check that a tag received from the
+// remote peer is valid for the given message and counter value.
+//
+// Parameters:
+//   verify      - If false, generate a new tag and write it to `tag`.
+//                 If true, verify the tag in `tag` against the message and
+//                 return 0 on success or a negative value on mismatch.
```
burgerdev: Please double-check all the mentions of "negative return" in this docstring. The function actually returns 1 or 0, exclusively, but I'd stick to "non-zero on failure" as with the other functions.

glome.h
```diff
@@ -11,32 +11,89 @@
...
+//                 buffer.
+//
+// Returns 0 on success, or a negative value if derivation fails.
 int glome_derive_key(const uint8_t private_key[GLOME_MAX_PRIVATE_KEY_LENGTH],
                      uint8_t public_key[GLOME_MAX_PUBLIC_KEY_LENGTH]);
 
-// Generates or verifies the GLOME tag for the message. Requires passing in the
-// private key of the local peer and the public key of the remote peer.
+// glome_tag generates or verifies a GLOME authentication tag for a message.
+// The tag is an HMAC-SHA256 truncated to GLOME_MAX_TAG_LENGTH bytes, computed
```
burgerdev: That's an implementation detail, this docstring should just say "computes a tag of size GLOME_MAX_TAG_LENGTH".

burgerdev (changes requested):
Thanks for improving our docs!

glome.h
```diff
@@ -47,4 +108,4 @@ int glome_tag(bool verify, unsigned char counter,
 }  // extern "C"
 #endif
 
-#endif  // GLOME_H_
+#endif  // GLOME_H_
```
burgerdev: Please add the trailing newline.

burgerdev (commented):
Thanks for the fixes, lgtm - we're almost there :)

FreezB11:
@burgerdev so i just put a new-line at the EOF 
and is there any update on the follow up work for this project 
i see a lot of old issue that is still there, any i can help with if possible?

burgerdev (approved):

